PR #1, opened 2021-04-27
Feature/add swift package manager

danial-rokt:
add swift package manager

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -9,11 +9,11 @@ let package = Package(
     products: [
         .library(
             name: "Rokt_Widget",
-            targets: ["Rokt-Widget"]),
+            targets: ["Rokt_Widget"]),
     ],
     dependencies: [],
     targets: [
-        .binaryTarget(name: "Rokt-Widget",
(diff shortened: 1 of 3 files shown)


nedarokt (approved):

Package.swift
```diff
@@ -0,0 +1,20 @@
...
+    platforms: [.iOS(.v9)],
+    products: [
+        .library(
+            name: "Rokt-Widget",
+            targets: ["Rokt_Widget"]),
+    ],
+    dependencies: [],
+    targets: [
+        .binaryTarget(name: "Rokt_Widget",
+                      url: "https://rokt-eng-us-west-2-mobile-sdk-artefacts.s3.amazonaws.com/ios/3.5.0-alpha.2487/Rokt_Widget.xcframework.zip",
```
DJingSeo, line 17: alpha?

DJingSeo (approved):

Package.swift
danial-rokt: Yes, It is alpha. We would wait for the feedback of the partner and would release the stable version 3.5.0 afterwards

PR #132, opened 2026-04-20
feat: add PaymentExtensionAttributeEnricher to detect extension registration

jamesnrokt:
## Background

For Shoppable Ads to be functional a payment extension needs to be registered

## What Has Changed

- Adds an attribute enricher to log when a Payment Extension has been registered

## How Has This Been Tested

- Local tests pass

## Notes

N/A

### DIFF
--- a/Sources/Rokt_Widget/Util/PaymentExtensionAttributeEnricher.swift
+++ b/Sources/Rokt_Widget/Util/PaymentExtensionAttributeEnricher.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+private let BE_IS_PAYMENT_EXTENSION_REGISTERED_KEY = "paymentExtensionRegistered"
+
+/// Enriches experience requests with whether a payment extension has been registered.
+///
+/// Downstream systems use this attribute to determine whether Shoppable Ads requiring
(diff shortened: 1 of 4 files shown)


Sources/Rokt_Widget/Util/PaymentExtensionAttributeEnricher.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+private let BE_IS_PAYMENT_EXTENSION_REGISTERED_KEY = "paymentExtensionRegistered"
```
denischilik, line 3: Could you please update naming pf constant
example `private let keyAdsExperienceType = "adsExperience"`

denischilik: The rest looks good to me

jamesnrokt: @denischilik This follows the standard naming conventions of the other constants of this type in the repo. `BE_` is traditionally what we've used to denote items specific to the backend rather than constants fully in our control

thomson-t (approved):

PR #145, opened 2026-04-27
feat: Check Shoppable feature flag depending on attribute

jamesnrokt:
## Background

- Currently selectPlacements will fire even when we know there are no Shoppable Ads for a partner

## What Has Changed

- Adds a check for a specific attribute and checks for extension registration and feature flag before calling a shoppable experience

## How Has This Been Tested

- All local tests pass

## Notes

N/A

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/RoktInternalImplementation.swift
+++ b/Sources/Rokt_Widget/RoktInternalImplementation.swift
@@ -758,6 +758,13 @@ class RoktInternalImplementation {
             trackingConsent = ATTrackingManager.trackingAuthorizationStatus.rawValue
         }
 
+        if attributes[keyAdsExperienceType] == valueAdsExperienceShoppable {
+            guard isShoppableFeatureFlagGatePassed(identifier: viewName, onFailure: composedEventHandler),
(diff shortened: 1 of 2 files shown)


Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -1090,6 +1084,28 @@ class RoktInternalImplementation {
         )
     }
 
+    private func isShoppableFeatureFlagGatePassed(identifier: String?, onFailure: ((RoktEvent) -> Void)?) -> Bool {
```
thomson-t: I think these function can be named better. Prefix `is` suggests that they are just boolean returns, but they have side effects. Maybe `validateShoppableAdsFeatureFlag` and `validateShoppableAdsPaymentExtension`?

jamesnrokt: Much better naming suggestion. I've updated it to match!

nickolas-dimitrakas (approved):

thomson-t (approved):

PR #169, opened 2026-05-21
test(pact): scaffold consumer-driven contracts against transactions-api v2

crisryantan:
## Background

This PR scaffolds consumer-driven [pact](https://docs.pact.io) contract testing for the iOS SDK against the unified `transactions-api` v2 surface (`/v2/sessions/offers`, `/v2/sessions/events`).

Some upfront context that will make the diff make more sense:

**Mobile isn't a consumer of `transactions-api` yet.** The SDK still calls `/rokt-mobile/v1/*`. The pact contract introduced here describes the **v2 surface** the SDK will satisfy when it migrates from v1 to v2. Until cutover, every pact gate is advisory — failures won't block merges and the contract effectively serves as a migration spec.

**`V2OffersClient` ships in the SDK from this PR, but is not called by any production code path.** It lives at `Sources/Rokt_Widget/Networking/V2OffersClient.swift` (internal visibility) alongside the existing `RoktNetWorkAPI` v1 implementation. Putting the client (and its wire model: `V2OffersRequest`, `V2OffersPage`, `V2OffersPrivacy`, `V2OffersChannel`, `V2OffersCustomer`, `V2OffersInput`) in `Sources/` rather than `Tests/` means the pact contract constrains the actual shipping SDK code, not a test-only stand-in — at v1→v2 migration there's no separate "lift" step that could silently introduce drift. The client is `internal`, accessed from the contract test via `@testable import Rokt_Widget`, so it doesn't widen the SDK's public API. At cutover, callers in the v1 networking path switch to invoking `V2OffersClient.fetchOffers(input:)`; the wire shape is already pinned by this pact contract.

**This PR is CI-neutral.** A follow-up PR will add a GitHub Actions workflow (`.github/workflows/pact-consumer.yml`) that runs the consumer test on a GitHub-provided `macos-latest` runner on PRs and, on push to `main`, publishes the pact JSON directly to the pact broker (publicly readable, auth-gated writes). The publish step is gated by a repository environment (`pact-publish`) requiring a `@ROKT/sdk-engineering` reviewer to expose `PACT_BROKER_USERNAME`/`PACT_BROKER_PASSWORD`, so untrusted PR commits cannot reach the publish path. PR-supplied code runs only on GitHub-provided runners; the only secret-bearing surface (the broker write endpoint) is gated by maintainer approval. The CI design has been reviewed with security.

**Architectural pattern aligned with the equivalent web SDK contracts.** The consumer spec is shaped to match the established pattern: the test provides domain inputs only; the client owns wire-shape construction (headers, body); the pact matchers describe the expected wire shape and act as the source of truth. Drift in the client's outgoing request will fail at the pact mock service — the test cannot silently agree with arbitrary client output.

## What has changed

- **`Sources/Rokt_Widget/Networking/V2OffersClient.swift`** — new file. `V2OffersClient` (internal struct) + wire-model structs (`V2OffersInput`, `V2OffersRequest`, `V2OffersPage`, `V2OffersPrivacy`, `V2OffersChannel`, `V2OffersCustomer`). Initializer takes the session-shaped values a production iOS service would resolve from session / config services (`accountId`, `authToken`, `sessionId`, `mpSessionId`, `mpid`, `sdkVersion`, `pageInstanceGuid`). The single public method `fetchOffers(input:)` takes a per-call `V2OffersInput` (request id, page identifier, page URL, customer email, attributes) and internally assembles every header and the full request body. Hardcoded iOS-specific values (`rokt-platform-type: "iOS"`, `rokt-integration-type: "msdk-ios"`, `Content-Type: application/json`, channel type `"msdk"`) live in the client only. Callers do not supply them.
- **`Package.swift`** — adds [PactSwift](https://github.com/surpher/PactSwift) `1.2.x` as a test-target dependency, plus a new `ContractTests` test target at `Tests/ContractTests/`. The new target is intentionally separate from `Rokt_WidgetTests` so PactSwift doesn't bloat normal unit-test runs. `ContractTests` depends on `Rokt_Widget` so it can `@testable import` `V2OffersClient`.
- **`Tests/ContractTests/V2OffersClientPactSpec.swift`** — one happy-path consumer pact interaction with provider state `a valid offers request is submitted`. Drives `V2OffersClient.fetchOffers(input:)` with realistic domain inputs. Matchers follow the conventions used for the equivalent web SDK contracts: `Matcher.SomethingLike` for runtime-variable values (account id, auth token, session ids, request id, page identifier, etc.), `Matcher.RegexLike` for the `page_type` enum, and **exact strings** for values that are hardcoded constants in `V2OffersClient` (`"iOS"`, `"msdk-ios"`, channel `"msdk"`, `Content-Type`). Exact-string matching ensures the consumer test fails if the client ever drifts (e.g., starts sending `"ios-mobile"` instead of `"iOS"`) — type-only matching wouldn't catch that.
- **`.gitignore`** — adds `pacts/`. The pact JSON artifact is generated by the contract test, picked up by the GH Actions workflow (in the follow-up PR), and published directly to the pact broker. It is never committed to this repo.

## How has this been tested

Run locally on macOS via Xcode (Cmd+U) against an iOS Simulator destination:

```
xcodebuild test \
  -scheme Rokt-Widget \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro' \
  -only-testing:ContractTests
```

Verified outcomes:

- `test_offersHappyPath_pageDetectionMatches` passes (~0.06s on iPhone 17 Pro / iOS 26.5).
- Pact JSON is generated by PactSwift inside the simulator sandbox.
- The JSON's `consumer.name = rokt-sdk-ios`, `provider.name = transactions-api`, one interaction at POST `/v2/sessions/offers`, all 7 expected headers and the full body shape. Variable values carry a `type` matcher; the four fixed-constant values (`rokt-platform-type: "iOS"`, `rokt-integration-type: "msdk-ios"`, `channel.type: "msdk"`, `Content-Type: "application/json"`) carry no matcher (exact-string assertion).
- `pactSpecification.version: "3.0.0"` — compatible with the existing pact broker.

The publish side lands with the follow-up PR (`.github/workflows/pact-consumer.yml` — direct broker publish, env-protected). Existing infrastructure produces the correct pact JSON locally; end-to-end validation through the broker lands once the broker's public-read change is deployed.

## Notes

- **Initial scaffolding only — one happy-path interaction.** A follow-up PR will add the events endpoint (`/v2/sessions/events`) and a minimal-payload offers variant (provider state `no page detection on offers request`), bringing the merged pact JSON to three interactions. Once both PRs land, the contract surface matches what the web SDK already publishes for the same provider.
- This PR adds shipping code (`V2OffersClient` + wire models in `Sources/`) but no production code path invokes it; existing v1 networking is unchanged. No required-checks are added.
- The follow-up GH Actions workflow publishes directly to the broker; no internal CI infrastructure executes any PR-supplied Swift code at any point.
- Provider states used here (`a valid offers request is submitted`) already exist in the provider-side fixture — no fixture changes required on the provider side.

## Checklist

- [ ] I have made corresponding changes to the documentation. (deferred — a docs section explaining the v1→v2 migration path for iOS engineers lands with the follow-up PR)
- [x] I have verified that CI completes successfully. (GH Actions checks on this repo. The Buildkite pipeline is no longer associated with this repo as of 2026-05-22 — the publish side moves to a GH-Actions-direct-publish design in the follow-up PR.)
- [ ] All insignificant commits have been squashed. (intentionally separate commits for the scaffold, PactSwift v1.2.1 API adaptation, swiftlint fixes, the architectural refactor to match the established pattern, and the response to review feedback — happy to squash on request)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,7 @@ Package.resolved
 vendor/
 Tests/Rokt_WidgetTests/**/*.swift.plist
 Tests/SizeReport/build/
+
+# Pact JSON artifacts (generated locally by ContractTests, published to broker
+# by CI, never committed — see rokt-pact-mobile-runner for the publish path).
+pacts/
--- a/Tests/ContractTests/V2OffersClient.swift
+++ b/Tests/ContractTests/V2OffersClient.swift
@@ -1,0 +1,121 @@
+import Foundation
+
+/// Test-only client for the v2 sessions offers endpoint on transactions-api.
+///
+/// Mobile is not yet a consumer of this surface — production code still calls
+/// `/rokt-mobile/v1/*`. This client exists so the consumer-driven pact spec can
+/// describe the v2 contract that mobile will satisfy at v1→v2 migration. Lift
+/// into `Sources/Rokt_Widget/Networking/` at cutover. See architecture plan in
+/// rokt-pact-mobile-runner.
+struct V2OffersClient {
+    let baseURL: URL
+    let urlSession: URLSession
+
+    init(baseURL: URL, urlSession: URLSession = .shared) {
+        self.baseURL = baseURL
+        self.urlSession = urlSession
+    }
+
+    struct Headers {
+        let accountId: String
+        let authorization: String
+        let platformType: String
+        let integrationType: String
+        let requestId: String
+        let pageInstanceGuid: String
+
+        static func iOSDefaults(
+            accountId: String,
+            authorization: String,
+            requestId: String,
+            pageInstanceGuid: String
+        ) -> Headers {
+            Headers(
+                accountId: accountId,
+                authorization: authorization,
+                platformType: "iOS",
+                integrationType: "msdk-ios",
+                requestId: requestId,
+                pageInstanceGuid: pageInstanceGuid
+            )
+        }
+    }
+
+    struct Request: Encodable {
+        let sessionId: String
+        let mpSessionId: String
+        let mpid: String
+        let page: Page
+        let privacy: Privacy
+        let channel: Channel
+        let customer: Customer
+        let attributes: [String: String]
+
+        enum CodingKeys: String, CodingKey {
+            case sessionId = "session_id"
+            case mpSessionId = "mp_session_id"
+            case mpid
+            case page
+            case privacy
+            case channel
+            case customer
+            case attributes
+        }
+
+        struct Page: Encodable {
+            let pageIdentifier: String
+            let url: String
+
+            enum CodingKeys: String, CodingKey {
+                case pageIdentifier = "page_identifier"
+                case url
+            }
+        }
+
+        struct Privacy: Encodable {
+            let doNotTrack: Bool
+            let gpcEnabled: Bool
+            let doNotShareOrSell: Bool
+
+            enum CodingKeys: String, CodingKey {
+                case doNotTrack = "do_not_track"
+                case gpcEnabled = "gpc_enabled"
+                case doNotShareOrSell = "do_not_share_or_sell"
+            }
+        }
+
+        struct Channel: Encodable {
+            let type: String
+            let sdkVersion: String
+
+            enum CodingKeys: String, CodingKey {
+                case type
+                case sdkVersion = "sdk_version"
+            }
+        }
+
+        struct Customer: Encodable {
+            let email: String
+        }
+    }
+
+    func sendOffers(headers: Headers, body: Request) async throws -> (Data, URLResponse) {
+        let url = baseURL
+            .appendingPathComponent("v2")
+            .appendingPathComponent("sessions")
+            .appendingPathComponent("offers")
+        var request = URLRequest(url: url)
+        request.httpMethod = "POST"
+        request.setValue(headers.accountId, forHTTPHeaderField: "rokt-account-id")
+        request.setValue(headers.authorization, forHTTPHeaderField: "Authorization")
+        request.setValue(headers.platformType, forHTTPHeaderField: "rokt-platform-type")
+        request.setValue(headers.integrationType, forHTTPHeaderField: "rokt-integration-type")
+        request.setValue(headers.requestId, forHTTPHeaderField: "x-request-id")
+        request.setValue(headers.pageInstanceGuid, forHTTPHeaderField: "rokt-page-instance-guid")
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+
+        request.httpBody = try JSONEncoder().encode(body)
+
+        return try await urlSession.data(for: request)
+    }
+}
--- a/Tests/ContractTests/V2OffersClient.swift
+++ b/Tests/ContractTests/V2OffersClient.swift
@@ -3,63 +3,92 @@ import Foundation
 /// Test-only client for the v2 sessions offers endpoint on transactions-api.
 ///
 /// Mobile is not yet a consumer of this surface — production code still calls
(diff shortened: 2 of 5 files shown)


Tests/ContractTests/V2OffersClient.swift
```diff
@@ -0,0 +1,150 @@
+import Foundation
+
+/// Test-only client for the v2 sessions offers endpoint on transactions-api.
```
thomson-t: Why is this test only?
If we don't have it in the SDK, it is not verifying anything. 
We don't have to consume it now, but the client itself and the data models can be there in the SDK.

Tests/ContractTests/V2OffersClientPactSpec.swift
```diff
@@ -0,0 +1,140 @@
...
+        Self.mockService
+            .uponReceiving("a v2 sessions offers request from rokt-sdk-ios for a configured page")
+            .given(ProviderState(description: "a valid offers request is submitted", params: [:]))
+            .withRequest(
+                method: .POST,
+                path: "/v2/sessions/offers",
+                headers: [
+                    "rokt-account-id": Matcher.SomethingLike("account-456"),
+                    "Authorization": Matcher.SomethingLike("Bearer session-token-abc"),
+                    "rokt-platform-type": Matcher.SomethingLike("iOS"),
```
thomson-t: I think `SomethingLike` is just verifying the type. Is it verifying the exact string?
These can be exact.

.gitignore
```diff
@@ -10,3 +10,7 @@ Package.resolved
 vendor/
 Tests/Rokt_WidgetTests/**/*.swift.plist
 Tests/SizeReport/build/
+
+# Pact JSON artifacts (generated locally by ContractTests, published to broker
+# by CI, never committed — see rokt-pact-mobile-runner for the publish path).
+pacts/
```
thomson-t, line 16: Not sure if we can run these as pure swift test as UIKit dependency is going to be a problem.
We may need to use xcodebuild and get the generated pacts from elsewhere.

Tests/ContractTests/V2OffersClient.swift
crisryantan: Good call — moved `V2OffersClient` (and the wire model structs) to `Sources/Rokt_Widget/Networking/V2OffersClient.swift` in 0196286. Internal visibility, accessed by the contract spec via `@testable import Rokt_Widget`. The pact now constrains real shipping SDK code, and at v1→v2 migration the existing v1 networking callers can switch over to invoking `V2OffersClient.fetchOffers(input:)` directly without a separate "lift" step that could silently introduce drift. PR description updated to reflect the change in scope.

Tests/ContractTests/V2OffersClientPactSpec.swift
crisryantan: You're right — `SomethingLike` only matches by type, so the client could drift to `"ios-mobile"` and still pass the consumer test, defeating the wire-shape check. Pinned as exact strings in 0196286:

- `rokt-platform-type: "iOS"`
- `rokt-integration-type: "msdk-ios"`
- `channel.type: "msdk"`

Verified in the generated pact JSON that these now carry no `matchingRules` entry (exact-string assertion). Kept `SomethingLike` on values that legitimately vary per call: account id, auth token, request id, session ids, page identifier, customer email, etc. Convention encoded in a doc comment at the top of the spec for future maintainers.

thomson-t (approved):

PR #170, opened 2026-05-22
ci(pact): add GitHub Actions workflow for consumer pact publish

crisryantan:
## Background

Follow-up to #169 (consumer pact test scaffold). #169 added `V2OffersClient` + the PactSwift contract spec but left CI out of scope. This PR wires up the workflow that runs the consumer test on PRs and, on push to `main`, publishes the resulting pact JSON to the pact broker.

Two-job design:

- **`consumer-test`** — runs on PRs touching the contract surface, plus on every push to `main`. Executes the PactSwift spec on `macos-latest` with Xcode 26.2; generates pact JSON via a small `PACT_OUTPUT_DIR` env-var indirection; uploads the JSON as a workflow artifact. **No secrets** are exposed; PR code from public forks runs safely here because the GitHub-hosted runner is fully isolated.
- **`publish`** — runs only on push to `main`, depends on `consumer-test`, gated by the `pact-publish` GitHub environment. The environment carries required-reviewer protection and main-branch-only deployment; broker write credentials are exposed only after a maintainer clicks approve. The publish itself uses the official `pactfoundation/pact-cli:latest` Docker image.

## What has changed

- **`Tests/ContractTests/V2OffersClientPactSpec.swift`** — `setUp` reads `PACT_OUTPUT_DIR` env var (defaults to `"pacts"` if unset; preserves local behavior). CI passes a host path via `xcodebuild`'s `TEST_RUNNER_` env-var propagation, so the generated pact JSON lands somewhere `actions/upload-artifact` can find it instead of disappearing into the simulator data container.
- **`.github/workflows/pact-consumer.yml`** — new file. Two-job workflow described above. Triggers only when the contract surface or the workflow itself changes (`Sources/Rokt_Widget/Networking/V2OffersClient.swift`, `Tests/ContractTests/**`, `Package.swift`, `.github/workflows/pact-consumer.yml`). SHA-pinned third-party actions throughout.

## Setup required before the `publish` job will run

The `publish` job will be skipped until the `pact-publish` GitHub environment is configured in **Settings → Environments**:

| Type | Name | Source |
|---|---|---|
| Variable | `PACT_BROKER_BASE_URL` | Rokt pact broker URL |
| Variable | `PACT_BROKER_USERNAME` | Broker write username |
| Secret | `PACT_BROKER_PASSWORD` | Broker write password (from AWS Secrets Manager) |

The environment should also be configured with:

- **Required reviewers**: `@ROKT/sdk-engineering` (or any maintainer set the team prefers)
- **Deployment branches**: `Selected branches → main only`

The `consumer-test` job runs immediately on PRs without any setup.

## How has this been tested

Locally on macOS via `xcodebuild` against an iOS Simulator destination:

```
TEST_RUNNER_PACT_OUTPUT_DIR=/tmp/pact-out xcodebuild test \
  -scheme Rokt-Widget \
  -destination 'platform=iOS Simulator,name=iPhone 17 Pro,OS=latest' \
  -only-testing:ContractTests
```

Verified outcomes:

- `test_offersHappyPath_pageDetectionMatches` passes (~0.06s).
- Pact JSON appears at `/tmp/pact-out/rokt-sdk-ios-transactions-api.json` — the host path, not the simulator data container.
- JSON shape sanity: `consumer = rokt-sdk-ios`, `provider = transactions-api`, 1 interaction at `POST /v2/sessions/offers`, pact spec `3.0.0`.

The workflow itself has not been smoke-tested end-to-end on GitHub Actions yet — that's the verification step once this is merged + the environment is configured.

## Notes

- **Draft until the `pact-publish` environment is provisioned.** Once it exists, this PR's CI will show what the `consumer-test` job looks like on GitHub-hosted runners; we can flip to ready-for-review then.
- The `publish` job uses `pactfoundation/pact-cli:latest` rather than a pinned tag. Acceptable because (a) it's published by the pact foundation and (b) the worst-case failure mode is the publish step failing, which is `soft_fail`-equivalent given the consumer test still gates on contract validity.
- Tags applied on publish: `--branch ${GITHUB_REF_NAME}` (so `main`), `--tag forward-looking` (per the v1→v2 migration plan — the contract is aspirational until the SDK actually switches to v2).
- Follow-up: docs section (`documentation/pact-v2-migration.md`) referenced by the workflow comments will land in a later PR (M8 in the rollout plan).

## Checklist

- [ ] I have made corresponding changes to the documentation. (deferred — migration doc lands in a follow-up PR per the rollout plan)
- [x] I have added tests that prove my fix is effective or that my feature works. (PactSwift consumer spec from #169 exercises the path; locally verified the new PACT_OUTPUT_DIR plumbing produces the JSON at the expected host path)
- [ ] I have verified that CI completes successfully. (consumer-test job will be observable once this PR runs CI; publish job depends on environment provisioning)
- [x] All insignificant commits have been squashed. (intentionally two commits: PACT_OUTPUT_DIR plumbing on the spec, then the workflow YAML — separate concerns, each independently revertable)

### DIFF
--- a/.github/workflows/pact-consumer.yml
+++ b/.github/workflows/pact-consumer.yml
@@ -156,12 +156,20 @@ jobs:
             exit 1
           fi
 
+          # Image pinned to manifest digest (not :latest). Renovate keeps it
+          # current via the `# renovate: datasource=docker depName=...` comment
+          # convention. Don't replace with :latest or a moving tag; this step
+          # runs with broker write credentials, so a compromised upstream image
+          # would have direct credential-exfiltration access. See security
+          # review note from 2026-05-27 (Sushant Adhikari) for the threat model.
+          # renovate: datasource=docker depName=pactfoundation/pact-cli
+          PACT_CLI_IMAGE="pactfoundation/pact-cli:1.5.0.4@sha256:4708c3d61157bff90082a483e4ff70541eb66f951eb3e218ea3e80a557abd170"
           docker run --rm \
             -v "${PWD}/pacts:/pacts" \
             -e PACT_BROKER_BASE_URL \
             -e PACT_BROKER_USERNAME \
             -e PACT_BROKER_PASSWORD \
-            pactfoundation/pact-cli:latest \
+            "${PACT_CLI_IMAGE}" \
             broker publish /pacts \
               --consumer-app-version "${GITHUB_SHA}" \
               --branch "${GITHUB_REF_NAME}" \
--- a/.github/workflows/pact-consumer.yml
+++ b/.github/workflows/pact-consumer.yml
@@ -18,17 +18,17 @@ name: Pact Consumer
 on:
   pull_request:
     paths:
-      - "Sources/Rokt_Widget/Networking/V2OffersClient.swift"
-      - "Tests/ContractTests/**"
-      - "Package.swift"
-      - ".github/workflows/pact-consumer.yml"
+      - Sources/Rokt_Widget/Networking/V2OffersClient.swift
+      - Tests/ContractTests/**
+      - Package.swift
+      - .github/workflows/pact-consumer.yml
   push:
     branches: [main]
     paths:
-      - "Sources/Rokt_Widget/Networking/V2OffersClient.swift"
-      - "Tests/ContractTests/**"
-      - "Package.swift"
-      - ".github/workflows/pact-consumer.yml"
+      - Sources/Rokt_Widget/Networking/V2OffersClient.swift
+      - Tests/ContractTests/**
+      - Package.swift
+      - .github/workflows/pact-consumer.yml
 
 permissions:
   contents: read
--- a/.github/workflows/pact-consumer.yml
+++ b/.github/workflows/pact-consumer.yml
@@ -2,16 +2,21 @@ name: Pact Consumer
 
 # Consumer-driven pact testing for the iOS SDK against `transactions-api` v2.
 #
-# Two jobs:
-#  - `consumer-test` runs on every PR + every push to `main`. It executes the
-#    PactSwift spec against the in-process mock server, generates a pact JSON,
-#    and uploads it as a workflow artifact. No Rokt secrets are exposed; PR code
-#    from public forks is safe to run here because the GitHub-provided runner is
-#    fully isolated from Rokt infrastructure.
-#  - `publish` runs only on push to `main` and is gated by the `pact-publish`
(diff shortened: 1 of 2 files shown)


.github/workflows/pact-consumer.yml
```diff
@@ -0,0 +1,126 @@
...
+    steps:
+      - name: Download pact JSON
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: pact-json
+          path: pacts/
+
+      - name: Publish to pact broker
+        env:
+          PACT_BROKER_BASE_URL: ${{ vars.PACT_BROKER_BASE_URL }}
```
mattbodle, line 105: How is this set?

crisryantan: Setup in https://github.com/ROKT/rokt-sdk-ios/settings/environments

BrandonStalnaker (dismissed):

thomson-t (changes requested):
How is it reaching the stage envs?

crisryantan:
> How is it reaching the stage envs?

Not reaching stage envs today. Currently intentional. Publish tags --tag "forward-looking" only, no --environment stage. iOS doesn't have a v2 production code path yet, so there's no "deployed to stage" event to record. When v2 ships, the SDK release pipeline will start calling pact-broker record-deployment --environment stage alongside release builds, and can-i-deploy will flip from advisory → enforce.sdk-web is in the same advisory state today on the provider side for the same reason.

.github/workflows/pact-consumer.yml
```diff
@@ -0,0 +1,175 @@
...
+#    a refactor to a shared networking helper that V2OffersClient calls would
+#    still run the pact spec. Uses macos-latest for the iOS Simulator. No
+#    Rokt secrets are exposed; fork-PR code is safe to run here.
+#  - `publish` runs only on push to `main` AND only when contract-relevant
+#    files actually changed (per `detect-changes`). It is gated by the
+#    `pact-publish` GitHub environment (required-reviewers + main-branch-only).
+#    Skipping publish on no-op merges saves maintainer approval clicks while
+#    keeping the consumer-test safety net unconditional.
+#
+# Reading: ../../documentation/pact-v2-migration.md (lands with a later PR).
```
thomson-t: Please add it when it actually lands

crisryantan: yeah its annoying. it's adding too much comments when it does new pushes to the branch. let me close this and open a new one once everything is cleaned up

PR #171, opened 2026-05-23
test(pact): add V2EventsClient + consumer pact for /v2/sessions/events

crisryantan:
## Summary

Adds a forward-looking consumer pact spec for the v2 RecordEvents endpoint (`POST /v2/sessions/events`), mirroring the existing `V2OffersClient` + `V2OffersClientPactSpec` pair (#169). The SDK does not yet call `/v2/sessions/events` in production — events still flow via `/rokt-mobile/v1/*` — so this pins the wire shape that mobile will satisfy at v1→v2 migration time.

## What ships

- **`Sources/Rokt_Widget/Networking/V2EventsClient.swift`** — owns wire-shape construction (headers + body) from domain inputs.
- **`Tests/ContractTests/V2EventsClientPactSpec.swift`** — consumer pact driving `V2EventsClient.recordEvents(events:)`. One happy-path interaction: single impression event, asserts the 202 response shape (`session_token`, `event_ids`, `errors`, `warnings`).

## Pattern parity with the offers spec

- `V2EventsClient` lives in `Sources/` rather than `Tests/` — pact constrains the shipping SDK code, not a test-only shadow.
- Wire-shape construction lives entirely in the client. The spec drives it with domain inputs only.
- Hardcoded constants (`"iOS"`, `"msdk-ios"`, `"msdk"`) are pinned exactly. Per-runtime values (account id, auth token, sdk version, event type, etc.) match by type via `SomethingLike`.

## Endpoint specifics (per `transactions/apps/api/internal/events/handler.go`)

- Session identity comes from the JWT in `Authorization`, not the body.
- `events[].timestamp` is int64 epoch millis (canonical format). RFC 3339 strings are still accepted but trigger a deprecation warning in the response.
- The 202 response always carries a refreshed `session_token` (the server may rotate it), per-event `event_ids`, and `errors` / `warnings` arrays.

## CI note

The `Consumer Test (PactSwift)` job won't appear on this PR's checks until #170 (which adds the `pact-consumer.yml` workflow) merges to `main`. Once #170 is on `main`, rebasing this PR will pick up the workflow and validate the new spec end-to-end.

## Test plan

- [ ] Merge #170 (workflow + path-filter removal).
- [ ] Rebase this PR onto `main`; confirm `Consumer Test (PactSwift)` runs and passes.
- [ ] Merge.
- [ ] Approve the `Publish to Broker` deployment on the resulting push-to-main; verify both `rokt-sdk-ios → transactions-api` pacts (offers + events) land on the broker via anonymous curl.

### DIFF
--- a/Sources/Rokt_Widget/Networking/V2EventsClient.swift
+++ b/Sources/Rokt_Widget/Networking/V2EventsClient.swift
@@ -1,0 +1,103 @@
+// periphery:ignore:all - referenced from Tests/ContractTests/V2EventsClientPactSpec.swift,
+// which isn't part of the rokt-Example scheme that Periphery scans. See V2OffersClient
+// for the rationale on why this lives in Sources/ rather than Tests/.
+
+import Foundation
+
+/// Client for the v2 sessions events endpoint on transactions-api.
+///
+/// Mirrors V2OffersClient — describes the v2 wire contract that mobile will
+/// satisfy at v1→v2 migration time. Not yet wired into production: the SDK
+/// still posts events via `/rokt-mobile/v1/*`. Lives in Sources/ so the
+/// consumer pact in `Tests/ContractTests/V2EventsClientPactSpec.swift`
+/// constrains the shipping SDK code rather than a test-only shadow.
+///
+/// Endpoint semantics (see transactions:apps/api/internal/events/handler.go):
+///   - Session identity is conveyed solely via the `Authorization` JWT;
+///     there is no session_id in the body.
+///   - `events[].timestamp` is Unix epoch milliseconds (int64). RFC 3339
+///     strings are accepted but deprecated and surface a warning on the
+///     202 response.
+///   - The response always includes a refreshed `session_token` (the
+///     server may rotate it), `event_ids` for each accepted event, and
+///     `errors` / `warnings` arrays (empty on the happy path).
+internal struct V2EventsClient {
+    let baseURL: URL
+    let accountId: String
+    let authToken: String
+    let sdkVersion: String
+    let urlSession: URLSession
+
+    init(
+        baseURL: URL,
+        accountId: String,
+        authToken: String,
+        sdkVersion: String,
+        urlSession: URLSession = .shared
+    ) {
+        self.baseURL = baseURL
+        self.accountId = accountId
+        self.authToken = authToken
+        self.sdkVersion = sdkVersion
+        self.urlSession = urlSession
+    }
+
+    func recordEvents(events: [V2Event]) async throws -> (Data, URLResponse) {
+        let url = baseURL
+            .appendingPathComponent("v2")
+            .appendingPathComponent("sessions")
+            .appendingPathComponent("events")
+
+        var request = URLRequest(url: url)
+        request.httpMethod = "POST"
+        request.setValue(accountId, forHTTPHeaderField: "rokt-account-id")
+        request.setValue(authToken, forHTTPHeaderField: "Authorization")
+        request.setValue("iOS", forHTTPHeaderField: "rokt-platform-type")
+        request.setValue("msdk-ios", forHTTPHeaderField: "rokt-integration-type")
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+
+        let body = V2EventsRequest(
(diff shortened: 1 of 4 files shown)


Sources/Rokt_Widget/Networking/V2EventsClient.swift
```diff
@@ -0,0 +1,103 @@
...
+    }
+}
+
+/// Single event in a v2 RecordEvents batch.
+///
+/// `timestamp` is int64 Unix epoch milliseconds — the canonical wire
+/// format. `data` is intentionally `[String: String]?` for now: the
+/// server-side schema is `map[string]any`, but iOS event payloads in
+/// practice carry only string values (e.g. `source_message_id`). Lift
+/// the value type when a real consumer needs nested or numeric data.
```
thomson-t: IMO, we don't need to add this level of verbose comments for everything. Some things are self explanatory. We can add comments when things are confusing or hard demands an explanation.

Sources/Rokt_Widget/Networking/V2EventsClient.swift
```diff
@@ -0,0 +1,103 @@
...
+
+/// Client for the v2 sessions events endpoint on transactions-api.
+///
+/// Mirrors V2OffersClient — describes the v2 wire contract that mobile will
+/// satisfy at v1→v2 migration time. Not yet wired into production: the SDK
+/// still posts events via `/rokt-mobile/v1/*`. Lives in Sources/ so the
+/// consumer pact in `Tests/ContractTests/V2EventsClientPactSpec.swift`
+/// constrains the shipping SDK code rather than a test-only shadow.
+///
+/// Endpoint semantics (see transactions:apps/api/internal/events/handler.go):
```
thomson-t: I don't think this a good idea to add comments about our internal system details.

Sources/Rokt_Widget/Networking/V2EventsClient.swift
```diff
@@ -0,0 +1,103 @@
...
+        request.setValue("msdk-ios", forHTTPHeaderField: "rokt-integration-type")
+        request.setValue("application/json", forHTTPHeaderField: "Content-Type")
+
+        let body = V2EventsRequest(
+            channel: V2EventsChannel(type: "msdk", sdkVersion: sdkVersion),
+            events: events
+        )
+        request.httpBody = try JSONEncoder().encode(body)
+
+        return try await urlSession.data(for: request)
```
thomson-t: Depending on how we consume the client, this will change. Currently, we use `RoktHTTPClient` for network requests to wrap the connection.

Sources/Rokt_Widget/Networking/V2EventsClient.swift
crisryantan: Done in 82ed2ed — dropped the doc block; the type signature is self-explanatory.

Sources/Rokt_Widget/Networking/V2EventsClient.swift
crisryantan: Done in 82ed2ed — removed the file-placement rationale and the endpoint semantics block. That context belongs in the PR description, not the source file.

Sources/Rokt_Widget/Networking/V2EventsClient.swift
crisryantan: Addressed in ef8c3da — both V2EventsClient and V2OffersClient now go through `HTTPClientAdapter` (default `RoktHTTPClient()`). Refactored both clients together for consistency rather than leaving one diverged.

Trade-offs from the original `URLSession`-only version:
- Async surface stays `async throws` via `withCheckedThrowingContinuation` around the existing callback API
- Encodable bodies → `RoktHTTPParameters` via `JSONEncoder` + `JSONSerialization.jsonObject(...)` (keeps type safety on inputs; small allocation cost)
- Return type changed from `(Data, URLResponse)` → `(Data?, HTTPURLResponse?)` to match what `RoktHTTPRequestResult` exposes
- Network errors throw; HTTP non-2xx returns normally so the pact verifier can assert any status code

Pact specs still pass locally (both interactions, 0.1s each).

BrandonStalnaker (approved):

PR #179, opened 2026-05-29
feat(ci): consumer pact publishing + advisory can-i-deploy

crisryantan:
## What

Adds the consumer-side Pact contract-testing workflow for the SDK:

- **`consumer-test`** runs the PactSwift contract specs in an iOS simulator and uploads the generated pact JSON as an artifact.
- **`publish`** posts the pact to the Pact broker on `main` pushes and same-repo (non-fork) PR pushes that touch contract-relevant paths, tagged with the source branch so per-branch pacts don't pollute the `main` contract.
- **`can-i-deploy`** queries the broker on PRs. Advisory for now (`continue-on-error: true`); becomes a blocking merge gate once the provider side is bootstrapped.

Broker connection details come entirely from repository/environment configuration (`vars`/`secrets` on the `pact-publish` environment) — nothing infrastructure-specific is hardcoded in the workflow.

## Notes

- Replaces the earlier **closed #170** (which used a manual approval gate). Publishing is now fully automated per team agreement; the environment is referenced only to scope the broker write credentials to the publish/can-i-deploy jobs.
- **Fork PRs** run the contract tests but skip `publish` (no credentials exposed).

## Validation

Validated end-to-end on a throwaway branch: the pact published successfully and is readable on the broker; `can-i-deploy` runs advisory (the provider-side `stage` environment isn't bootstrapped yet, so it reports not-deployable without blocking).

## How has this been tested?

- **Contract specs** (`ContractTests`, PactSwift) run in CI on every PR and generate the pact JSON.
- **Publish path** validated end-to-end via a throwaway smoke PR — #180 — which ran the `Publish to Broker` job from CI and confirmed the pact publishes successfully to the broker. The smoke branch was deleted afterward; its run history remains on #180 for reference.

### DIFF
--- a/.github/workflows/pact-consumer.yml
+++ b/.github/workflows/pact-consumer.yml
@@ -136,13 +136,15 @@ jobs:
               --build-url "${BUILD_URL}"
 
   can-i-deploy:
-    name: Can I Deploy
+    name: Can I Deploy (advisory)
     needs: [publish]
     if: github.event_name == 'pull_request'
     runs-on: ubuntu-latest
-    # Advisory until the provider side is bootstrapped for environment-based
-    # can-i-deploy; flip to false to make this a blocking merge gate.
-    continue-on-error: true
+    # Advisory: the published pacts are forward-looking V2 contracts the SDK
+    # does not consume yet, so there is no `stage` deployment to verify
+    # against — the step reports the broker's answer but never fails the job.
+    # When the SDK starts consuming the V2 APIs, make this a blocking gate
+    # (remove the advisory wrapper in the step below) and re-validate end-to-end.
     environment: pact-publish
     steps:
       - name: Check provider verification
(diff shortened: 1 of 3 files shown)


.github/workflows/pact-consumer.yml
```diff
@@ -0,0 +1,172 @@
...
+          if docker run --rm \
+            -e PACT_BROKER_BASE_URL \
+            -e PACT_BROKER_USERNAME \
+            -e PACT_BROKER_PASSWORD \
+            "${PACT_CLI_IMAGE}" \
+            broker can-i-deploy \
+              --pacticipant rokt-sdk-ios \
+              --version "${GITHUB_SHA}" \
+              --to-environment stage; then
+            echo "can-i-deploy: deployable."
```
thomson-t, line 168: There is a `400` error returned from the [pact](https://github.com/ROKT/rokt-sdk-ios/actions/runs/26673637123/job/78621564736?pr=179#step:2:110). 
Is this expected?

crisryantan: Yes.. this is expected and harmless. The reason is that the step is advisory, so it reports whatever the broker says but never actually fails the job. That's why you see the red `400` line sitting next to a green check.

The error itself is `Environment 'stage' does not exist`. We haven't set up any deployment environments in the broker yet, so any `--to-environment` call will return that, and it isn't specific to `stage`. We're not gating on it right now because these are forward-looking contracts(This would be @jaissica12 and @khushi1033's work to satisfy the contract once v2 API is consumed). The SDK publishes its v2 pact, but it doesn't actually call the v2 APIs in shipped code yet, so there's nothing deployed to verify against. So `stage` here is really just a placeholder for now.

Once the SDK starts consuming v2, we'll turn this into a blocking gate and point it at the environment a released SDK actually calls, most likely `production`.

thomson-t (approved):

PR #182, opened 2026-06-01
feat: Convert to Monorepo

BrandonStalnaker:
## Background

This work turns the iOS SDK repo into a monorepo so the widget SDK and related Swift packages can be developed and versioned together in one place, while still mirroring those packages to their standalone GitHub repos on release. It also aligns tooling: first-party code under Packages/ is linted and formatted with the same Trunk / SwiftLint / SwiftFormat / objc checks as the rest of the tree, instead of excluding subtree copies or relying on per-package configs that never ran in this repo.

Motivation / context for reviewers
- Subtrees under Packages/ are the source of truth here; release automation uses Packages/matrix.json and the publish workflow to bump versions, mirror, and publish in a defined order.
- SwiftLint only merges nested configs (e.g. Packages/*/Tests/.swiftlint.yml) when SwiftLint is not invoked with --config; Trunk is configured accordingly so trunk check matches local expectations.
- Optional stored properties on @objc classes cannot always be marked @objc; the objc helper script was tightened so we do not require @objc on optional-typed public let / public var lines.
- Checkov was flagging anonymized test JSON; those paths are skipped for Checkov in Trunk because -f scans do not honor skip-path the same way.

## What Has Changed

- Monorepo layout: Added subtree packages under Packages/ (e.g. UX helper and payment extension), with Packages/matrix.json describing mirror targets, CocoaPods order, and Trunk metadata for release.
- Versioning & release: Updated Release – Draft and Release – Publish workflows for a single ecosystem version bump, matrix-driven mirror/publish steps, and kits-path: Packages (input value; path is Packages/, not Kits/).
- SPM & CocoaPods: Root Package.swift / Rokt-Widget.podspec and related scripts (e.g. scripts/pod_push.sh) updated for local package paths and publishing behavior documented in MONOREPO.md.
- Docs: MONOREPO.md, AGENTS.md, and README updated for the monorepo workflow, validation commands, and how Packages/ is linted (including nested SwiftLint under package Tests and manual swiftlint without --config when Trunk is unavailable).
- Lint / format: Root .swiftlint.yml includes Packages/; removed duplicate package-root .swiftlint.yml files; added Packages/*/Tests/.swiftlint.yml overlays for test-only rule relaxations; .trunk/trunk.yaml sets SwiftLint direct_configs: [] so nested configs merge; Trunk Checkov ignores for selected test/fixture JSON paths.
- Swift / quality: Small fixes in package Sources (e.g. SwiftLint nesting handling, RoktPluginViewState @objc only where valid); objc-prop-check script updated to exempt optional-typed public stored properties on @objc public classes.

## Checklist


nickolas-dimitrakas (dismissed):
LGTM, maybe make note of the zizmor comments so we can address at a later date.

nickolas-dimitrakas (approved):

PR #183, opened 2026-06-02
feat(contract): add v2/sessions/init consumer pact for iOS SDK

crisryantan:
## What

Adds consumer-driven contract coverage for `POST /v2/sessions/init` to the iOS SDK:

- **`Sources/Rokt_Widget/Networking/V2SessionsInitClient.swift`** — new networking client that owns the wire-shape construction (headers + JSON body), following the existing `V2OffersClient` / `V2EventsClient` pattern. Public method:
  `initSession(operating_system:sdk_version:layout_schema_version:) async throws -> (Data?, HTTPURLResponse?)`.
- **`Tests/ContractTests/V2SessionsInitClientPactSpec.swift`** — a PactSwift consumer spec (`test_sessionInitHappyPath_returnsSessionAndFeatureFlags`) that drives the client through the pact mock service. Because the test only supplies domain inputs and never builds headers/body itself, any drift in the wire shape fails the test.
- **`.github/workflows/pact-consumer.yml`** — registers the new client in the contract paths filter so future changes to it re-trigger the consumer job.

## Contract shape

- **Request** — `POST /v2/sessions/init`, body `{ operating_system, sdk_version, layout_schema_version }`. The enum value `"ios"` and the `rokt-platform-type` / `rokt-integration-type` headers are pinned as exact strings (catches client drift); per-call headers (account id, auth token, request id) use a regex `.+` matcher and per-call body values use type matchers — matching the shared contract and the Android consumer.
- **Response** — HTTP 200 asserting `session_id`, `session_token` (`token` + `expires_at`) and the `feature_flags` map, all with type matchers (server-evaluated flags vary by environment). `fonts` is pinned as an exact empty array, since the provider always returns `[]` and the item shape is not yet frozen.

## Rollout

Part of a three-repo rollout — transactions provider + iOS and Android consumers. Pacts publish under the **forward-looking** tag: the SDK does not yet call `/v2/sessions/init` in shipped code, so provider verification is advisory (soft-fail) until both mobile SDKs consume v2 in production. Provider-state string and interaction description are coordinated byte-for-byte across all three repos.

## Verification

- `xcodebuild test -only-testing:ContractTests` — all three specs (Offers, Events, SessionsInit) pass locally.
- Generated pact confirmed: `POST /v2/sessions/init`, correct provider state, 200 response with the asserted fields and `fonts: []`.
- `trunk check` clean on the new files (swiftlint + swiftformat).

### DIFF
--- a/.github/workflows/pact-consumer.yml
+++ b/.github/workflows/pact-consumer.yml
@@ -34,6 +34,7 @@ jobs:
             contract:
               - 'Sources/Rokt_Widget/Networking/V2OffersClient.swift'
               - 'Sources/Rokt_Widget/Networking/V2EventsClient.swift'
+              - 'Sources/Rokt_Widget/Networking/V2SessionsInitClient.swift'
               - 'Tests/ContractTests/**'
               - 'Package.swift'
               - '.github/workflows/pact-consumer.yml'
--- a/Sources/Rokt_Widget/Networking/V2SessionsInitClient.swift
+++ b/Sources/Rokt_Widget/Networking/V2SessionsInitClient.swift
@@ -1,0 +1,91 @@
+// periphery:ignore:all - referenced from Tests/ContractTests/V2SessionsInitClientPactSpec.swift
+
+import Foundation
+
+internal struct V2SessionsInitClient {
+    let baseURL: URL
+    let accountId: String
+    let authToken: String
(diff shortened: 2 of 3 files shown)


nickolas-dimitrakas (approved):
LGTM

PR #185, opened 2026-06-03
feat: v2 init config, models, and session store

khushi1033:
## Background

Lands the v2 init enablement work in the iOS SDK: the foundation for talking to `/v2/sessions/init` (response models, session store) plus the coordinator that ties them together. All net-new files, **not yet wired into any live path** (`Rokt.initialize()` is unchanged) — this builds the v2 path in parallel so it can be switched on in a later PR.

New files carry `// periphery:ignore:all` since they're unreferenced from production until the wiring PR. Reuses the v2 sessions init client + Pact spec from #183.

## What Has Changed

- **`Environment.gatewayBaseURL`** — gateway base URL per environment, added as an extension on the existing `Environment` type so the v2 path reuses the same environment without a parallel config type.
- **`TxnInitResponse` models** — `Decodable` types for the v2 response (`TxnInitResponse`, `TxnSessionToken`, `TxnFontItem`, `TxnFeatureFlags`), plus a mapper into the existing `InitFeatureFlags`. Handles the mixed-type flag map and treats `expires_at` as epoch millis (the JWT is never decoded). Font-related feature flags are intentionally excluded from the v2 mapping as they are unused.
- **`TxnSessionManager`** — thread-safe store for the session id + token. Exposes a `Bearer` auth header only while the token is present and unexpired (`nil` otherwise → server mints a fresh session, no client-side 401 handling). Reusable beyond init via a token-only refresh.
- **`TxnInitService`** — coordinator that resolves the gateway base URL, sends the init request, decodes the response, updates the session store, and returns the mapped feature flags. Adds exponential-backoff-with-jitter retry on transient failures (timeouts / 5xx / connection errors; fails fast on offline, 4xx, and decode errors).
- **`V2SessionsInitClient`** — auth token is now optional, so a missing/expired token omits the `Authorization` header (server mints a fresh session). Existing Pact spec stays green.
- Unit tests for all of the above.

## Design decisions

- **Session store is in-memory at this stage** — the v2 token is short-lived and refreshed on every response, so cross-launch persistence was deferred. The wiring PR (#192) adds Keychain persistence for session-reuse parity with Android.
- **No central endpoints file** — each v2 client owns its own path from the gateway base URL (matching the existing v2 clients), so config exposes only the base URL.
- **Retry lives in the coordinator, not the client** — the v2 client calls the HTTP layer directly (no shared retry), so `TxnInitService` re-adds the resilience the legacy path had, with a self-contained retry predicate rather than reusing legacy networking helpers.

## How Has This Been Tested

- Unit tests run on the iOS Simulator (`xcodebuild test -scheme Rokt-Widget`) — all green. Coordinator coverage includes the success path, `Authorization` header present-with-token / omitted-without, retry-then-succeed on 5xx and timeout, retry exhaustion, and fast-fail on offline / 4xx / decode error.
- `trunk check` and `periphery scan` clean.
- No live-path behavior to verify; the code isn't reachable from the public API yet.

## Notes

## Checklist


Sources/Rokt_Widget/Constants/TxnConfiguration.swift
```diff
@@ -0,0 +1,31 @@
+// periphery:ignore:all - net-new v2 init configuration, not yet wired into the live path
+import Foundation
+
+internal enum TxnEnvironment: Equatable {
```
thomson-t: Do we need this? 
Wy don't we use the existing environment?

Sources/Rokt_Widget/Models/TxnInitResponse.swift
```diff
@@ -0,0 +1,168 @@
...
+                items[key] = FeatureFlagItem(match: !version.isEmpty)
+            case .int, .double:
+                continue
+            }
+        }
+
+        return InitFeatureFlags(
+            roktTrackingStatus: bool(forKey: "rokt-tracking-status") ?? true,
+            shouldLogFontHappyPath: bool(forKey: "ios-sdk-log-font-happy-path") ?? false,
+            shouldUseFontRegisterWithUrl: bool(forKey: "ios-sdk-use-font-register-with-url") ?? false,
```
thomson-t: I think we can retire these feature flags.

khushi1033: Yep double checked and both these are not used - updated the code so we're not pulling them into v2

Sources/Rokt_Widget/Constants/TxnConfiguration.swift
khushi1033: Good point. The idea was to add a new environment since we were leaning towards all net new code, but using the existing one makes more sense and matches what is done on the android side. Updated to use the existing environment.

thomson-t (approved):

Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
```diff
@@ -0,0 +1,65 @@
+// periphery:ignore:all - net-new v2 session store, not yet wired into the live path
+import Foundation
+
+internal final class TxnSessionManager {
```
denischilik, line 4: could you please use actor here it will remove all locks from the class

Sources/Rokt_Widget/Models/TxnInitResponse.swift
```diff
@@ -0,0 +1,166 @@
...
+    func toInitFeatureFlags() -> InitFeatureFlags {
+        var items: [String: FeatureFlagItem] = [:]
+        for (key, value) in flags {
+            switch value {
+            case .bool(let flag):
+                items[key] = FeatureFlagItem(match: flag)
+            case .string(let version):
+                items[key] = FeatureFlagItem(match: !version.isEmpty)
+            case .int, .double:
+                continue
```
denischilik, line 157: should we skip `.int, .double`
i'm asking because we parse them here
```
        } else if let value = try? container.decode(Int.self) {
            self = .int(value)
        } else if let value = try? container.decode(Double.self) {
            self = .double(value)
```

Sources/Rokt_Widget/Models/TxnInitResponse.swift
```diff
@@ -0,0 +1,166 @@
+// periphery:ignore:all - net-new v2 init response models, not yet wired into the live path
+import Foundation
+
+internal struct TxnInitResponse: Decodable, Equatable {
+    let sessionId: String
+    let sessionToken: TxnSessionToken
+    let featureFlags: TxnFeatureFlags
```
denischilik, line 7: consider to make `TxnSessionToken` and `TxnFeatureFlags` as embeded types of `TxnInitResponse`

Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
khushi1033: Got it, will make a follow-up PR to do this

Sources/Rokt_Widget/Models/TxnInitResponse.swift
khushi1033: Got it, we're planning to make a PR to refactor and remove the TxN naming convention, so will couple this change with it

Sources/Rokt_Widget/Models/TxnInitResponse.swift
khushi1033: We do need them, looks like a no-op here, but there are numeric consumers in the wiring that do use it ([init wiring PR](https://github.com/ROKT/rokt-sdk-ios/pull/192))

PR #191, opened 2026-06-04
feat: Offer Retry on PayPal Cancel or Payment Failure

BrandonStalnaker:
## Background

Built-in PayPal for device pay / shoppable flows opens the processor approval URL in SFSafariViewController. When the buyer left that step without completing approval (Safari dismissed or cancel redirect), the SDK treated it like a finished checkout: it invoked the deferred processPayment completion and the forward-payment completion path, which drove devicePayFinalized and forwardPaymentFinalized with failure semantics. That tore down the confirmation UI and cleared instant-purchase state even though the user had only backed out of the browser step. The pending PayPal checkout snapshot was also consumed on the first forward-payment presentation, so a second tap could not reopen the same hosted approval without starting over.

## What Has Changed

This change keeps the forward-payment confirmation experience usable after a hosted cancel: on .canceled we do not notify UX finalization for that attempt, do not fire the deferred Step-1 completion, and re-queue the same pending PayPal snapshot so the confirm action can launch approval again. Success and real orchestration failures are unchanged.

## How Has This Been Tested

Tested on simulator with spoofed API resposne

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/Models/Payment/PaymentOrchestrator.swift
+++ b/Sources/Rokt_Widget/Models/Payment/PaymentOrchestrator.swift
@@ -370,6 +370,11 @@ final class PaymentOrchestrator {
     /// backend webhook finalizes the purchase from the PayPal redirect; the SDK only needs
     /// to surface success/cancel/failure to the UXHelper.
     ///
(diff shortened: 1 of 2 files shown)


thomson-t (approved):

jamesnrokt (approved):
Looks good. Please could we attach a video of the retry flow?

PR #192, opened 2026-06-08
feat: wire up txn init endpoint

khushi1033:
## Background

The v2 Transactions init components (init client/service, response model, session manager) were added in the previous PR but not yet used. This PR wires them into the live SDK init flow so initialization now calls the v2 `sessions/init` endpoint and maps the result into the existing init model, leaving all downstream behavior unchanged. A single switch retains the legacy path for rollback.

Depends on the v2 init foundation PR.

## What Has Changed

- Init now calls the v2 `sessions/init` endpoint; the response is adapted into the existing `InitRespose` so timeouts, fonts, and feature-flag consumers are unchanged.
- The v2 init request runs through the SDK's shared networking client, so it inherits the configured request timeout and cache policy.
- Added a single rollback switch to fall back to the legacy init path if needed.
- Session token is persisted securely in the Keychain across launches (device-only, available after first unlock), bound to the current tag id, with expired tokens dropped on load.
- Added an offline mock transport + bundled fixture so the example app/playground exercises the real v2 decode + mapping path with no network.
- Scoped the secret scanner to skip example test JSON fixtures.

## How Has This Been Tested

- New unit tests for the response adapter, session persistence, mock transport, and the init wiring (success + failure paths).
- UI test harness now stubs the v2 `sessions/init` endpoint, so the existing UI suite exercises the v2 init path.
- `trunk check` clean; `rokt-Example` builds; unit + UI suites green.
- Manual Stage run via the example app: confirmed the v2 `sessions/init` request fires and init completes successfully.

## Checklist


Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
```diff
@@ -56,10 +86,45 @@ internal final class TxnSessionManager {
...
+        // Drop a persisted-but-expired token so we never start with stale state;
+        // an expired JWT is dead server-side and a fresh session is minted at init.
+        if isExpiredLocked {
+            clear()
+        }
+    }
+
+    private func persist(includeSessionId: Bool) {
+        guard let userDefaults, let roktTagId else { return }
+        if includeSessionId {
```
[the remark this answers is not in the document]
khushi1033: Addressed - added `KeychainTxnSessionStore` that reads/writes via the Keychain instead of UserDefaults

Sources/Rokt_Widget/Networking/MockTxnInitHTTPClient.swift
```diff
@@ -0,0 +1,68 @@
...
+        completionQueue: DispatchQueue,
+        completionHandler: ((RoktHTTPRequestResult) -> Void)?
+    ) -> URLRequest? {
+        onRequestStart?()
+
+        let url = URL(string: urlAddress) ?? URL(string: TxnEnvironment.prod.gatewayBaseURL)!
+        let result: RoktHTTPRequestResult
+
+        if let fixtureURL = bundle.url(forResource: Self.fixtureName, withExtension: "json"),
+           let data = try? Data(contentsOf: fixtureURL) {
```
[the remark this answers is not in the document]
khushi1033: Fixed - `MockTxnInitHTTPClient` now falls back to a built-in success response, so Mock init no longer depends on the host bundling `txn_init.json`

Sources/Rokt_Widget/Constants/TxnConfiguration.swift
```diff
@@ -36,4 +36,15 @@ internal struct TxnConfiguration {
         default: return .prod
         }
     }
+
+    // Maps the live config.environment onto the v2 gateway. Mock reuses prod (the mock
+    // path swaps the transport, not the host); ProdDemo has no v2 host and reuses prod.
+    static func getEnvironment(_ environment: Environment) -> TxnEnvironment {
+        switch environment {
+        case .Stage: return .stage
+        case .Prod, .ProdDemo, .Mock: return .prod
```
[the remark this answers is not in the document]
khushi1033: Fixed - ProdDemo now stays on the legacy init path instead of falling through to prod, so demo integrations keep hitting the demo host. The v2 path covers the other environments.

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -45,6 +47,20 @@ class RoktInternalImplementation {
     var stateManager: StateBagManaging = StateBagManager()
     private var clientTimeoutMilliseconds: Double = RoktInternalImplementation.defaultTimeout
     private var defaultLaunchDelayMilliseconds: Double = RoktInternalImplementation.defaultDelay
+
+    // MARK: - v2 Transactions init (isolated)
+
+    // Single rollback switch: set to false to restore the legacy init path verbatim.
+    private static let useV2Init = true
```
thomson-t, line 56: Do we need this switch?
We will keep it in feature branch until it is ready and won't have the legacy calls when we release it, right?

khushi1033: Yes, the motivation for keeping it is without the flag, v1 init becomes unreachable dead code and Periphery checks would flag it. That would force us to either add `periphery:ignore` comments throughout v1 or delete v1 now, which we want to take on at this stage. Though happy to go the `periphery:ignore` route instead if you prefer

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -841,47 +857,115 @@ class RoktInternalImplementation {
...
+            layoutSchemaVersion: Self.txnLayoutSchemaVersion,
+            sessionManager: resolveTxnSessionManager(roktTagId: roktTagId),
+            httpClient: httpClient
+        )
+    }
+
+    // Reuse a single session manager per tag id so the minted bearer token is
+    // available to subsequent v2 calls (and is restored from storage on launch).
+    private func resolveTxnSessionManager(roktTagId: String) -> TxnSessionManager {
+        if let existing = txnSessionManager, existing.boundTagId == roktTagId {
```
thomson-t: Not sure about this logic. We would only ever support single accountId/TagId at a time. 
If we switch account, that means we should be clearing all the session and token information we have.

khushi1033: The clearing does happen, it's just implicit. When the tag id changes, `resolveTxnSessionManager` creates a fresh manager, and on init it checks whether the stored tag id matches the new one. Since it doesn't, it wipes the stored session, token, and expiry from the Keychain. So switching accounts clears everything, same as the legacy `SessionManager` already does.

The reason we key by tag id rather than using a single tag-less manager is the cross-launch case: when the app restarts we need to know whether the persisted token belongs to the current account. 

Let me know if that sounds right

Sources/Rokt_Widget/RoktInternalImplementation.swift
thomson-t: In that case, can we add a TODO item so that we can clean it up when we switch over?

Sources/Rokt_Widget/RoktInternalImplementation.swift
thomson-t: The concern is about the need of `resolveTxnSessionManager` itself. Why would we need to resolver the session manager if there is only ever going to one sessionmanager that is bound to the active tagId?

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -841,47 +857,115 @@ class RoktInternalImplementation {
...
+        guard Self.useV2Init else {
+            RoktAPIHelper.initialize(
+                roktTagId: roktTagId,
+                success: { self.handleInitSuccess($0, initStartTime: initStartTime) },
+                failure: { self.handleInitFailure(error: $0, statusCode: $1, response: $2) }
+            )
+            return
+        }
+
+        let service = makeTxnInitService?(roktTagId) ?? defaultTxnInitService(roktTagId: roktTagId)
```
thomson-t: I think when an init call happens it should create a new instance of session manager anyway. Only extra thing I would expect it to do is, load the persisted session token if there was one.

Sources/Rokt_Widget/RoktInternalImplementation.swift
khushi1033: Done added a TODO

Sources/Rokt_Widget/RoktInternalImplementation.swift
khushi1033: Ah got it. Makes sense to simplify it. Removed `resolveTxnSessionManager`, so `defaultTxnInitService` just creates a fresh `TxnSessionManager(roktTagId:)` directly on each init call. The manager's initializer handles the two cases: same tag as what's stored → restores the persisted session; different tag → clears it

Sources/Rokt_Widget/RoktInternalImplementation.swift
khushi1033: Yup thats what it does now, a new `TxnSessionManager` is created on each init call, and its initializer loads the persisted session token if one exists for that tag 

`makeTxnInitService` is test-only and nil in production. Renamed to `makeTxnInitServiceOverride` with a comment to make it more clear.

Sources/Rokt_Widget/DataAccess/TxnSessionStore.swift
```diff
@@ -0,0 +1,58 @@
...
+// fake in tests, so TxnSessionManager makes no Keychain calls under test.
+internal protocol TxnSessionStore: AnyObject {
+    func string(forKey key: String) -> String?
+    func setString(_ value: String, forKey key: String)
+    func removeValue(forKey key: String)
+}
+
+// The bearer token is a high-value credential, so it is encrypted at rest in the
+// Keychain rather than left in UserDefaults. Items are device-only and available
+// after first unlock so a background relaunch can still restore the session.
```
thomson-t: Is this token something that should have this level of security? 
Considering there is an expiry associated with it, would it qualify for encrypted at rest?
@jamesnrokt

khushi1033: Synced with the team and you're right we don't need to encrypt the token it in the keychain, its not done on the other platforms as well. Updated to use UserDefaults

Sources/Rokt_Widget/DataAccess/TxnSessionStore.swift
```diff
@@ -0,0 +1,30 @@
...
+
+    init(defaults: UserDefaults = .standard) {
+        self.defaults = defaults
+    }
+
+    func string(forKey key: String) -> String? {
+        defaults.string(forKey: key)
+    }
+
+    func setString(_ value: String, forKey key: String) {
```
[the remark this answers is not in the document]
khushi1033: Since the token expires pretty soon, no need for this level of security, its not encrypted on the other platforms either, so better to keep it simple and match those patterns.

Sources/Rokt_Widget/Networking/V2SessionsInitClient.swift
```diff
@@ -79,8 +77,11 @@ internal enum V2SessionsInitClientError: Error {
 }
 
 internal struct V2SessionsInitRequest: Encodable {
```
crisryantan, line 79: Let's get rid of the V2 suffix

Tests/Rokt_WidgetTests/TestTxnInitWiring.swift
```diff
@@ -0,0 +1,136 @@
+import XCTest
+@testable import Rokt_Widget
+
+final class TestTxnInitWiring: XCTestCase {
```
crisryantan, line 4: Have we also wired the implementation to run against the existing init pact tests?

Tests/Rokt_WidgetTests/Shared/DataAccess/TestTxnSessionManager.swift
```diff
@@ -76,4 +76,102 @@ final class TestTxnSessionManager: XCTestCase {
         XCTAssertNil(manager.authorizationHeader)
         XCTAssertTrue(manager.isExpired)
     }
+
```
crisryantan, line 79: From android side, we are going away with the prefixed Txn and V2 in our filenames. We however our folder structure and place these files under a folder named txn. See the #1039 in android PR for reference

khushi1033: Yep, planning to do a followup PR to handle renaming and moving files under one TxN directory. Didn't want do it here, since it would mix the scope of the PR

Sources/Rokt_Widget/Networking/V2SessionsInitClient.swift
khushi1033: Yeah, similar to the txn naming clean up, will do a followup to remove mentions of v2

Tests/Rokt_WidgetTests/TestTxnInitWiring.swift
khushi1033: Yep! `V2SessionsInitClient` is run against the existing pact, so the request shape, headers, and body are all contract-validated.

thomson-t (approved):

PR #194, opened 2026-06-10
feat: Add Card Forwarding Retry Support

BrandonStalnaker:
## Background

Built-in card forward payment popped Step‑1 state as soon as the user tapped confirm, so any /v1/cart/purchase outcome immediately drove Step‑1 completion and forwardPaymentFinalized. There was no way to stay on the confirm step and retry after transient failures. This PR adds an in‑flight state and client‑side retry vs terminal handling so a retryable error can return the session to “ready to confirm” without finalizing the flow. No new dependencies. Validate confirm UI with the pinned RoktUXHelper when forwardPaymentFinalized is omitted on retryable errors.

## What Has Changed

- PaymentOrchestrator: Pattern B — new cardInFlight case; beginBuiltInCardForwardPaymentIfReady(), isBuiltInCardForwardPaymentInFlight(), restoreBuiltInCardForwardPaymentAfterRetryableFailure(), finishBuiltInCardForwardPaymentAttempt(result:); removed popPendingBuiltInCardCompletion(); cancelPendingBuiltInTwoStepIfNeeded() treats cardInFlight like idle card pending.
- handleForwardPayment: Blocks a second POST while card forward is in flight; on retryable HTTP / business failures → restore only (no Step‑1 completion, no forwardPaymentFinalized); terminal paths unchanged (finish + finalize).

## How Has This Been Tested

Orchestrator tests updated for begin / finish / restore and PayPal coexistence.

## Notes

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/RoktInternalImplementation.swift
+++ b/Sources/Rokt_Widget/RoktInternalImplementation.swift
@@ -682,6 +682,48 @@ class RoktInternalImplementation {
         return (false, failureReason)
     }
 
+    /// Conservative client-side policy for when the buyer may retry built-in card forward-payment
+    /// without leaving the confirm step (HTTP 200 with ``PurchaseResponse/success`` == `false`).
+    private static func isRetryableForwardPaymentBusinessFailure(failureReason: String?) -> Bool {
+        let normalized = failureReason?
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+            .lowercased() ?? ""
+        if normalized.isEmpty { return false }
+        let hints = [
+            "timeout",
+            "timed out",
+            "try again",
+            "temporarily unavailable",
+            "service unavailable",
+            "too many requests",
+            "throttle",
+            "unavailable"
+        ]
+        return hints.contains { normalized.contains($0) }
+    }
+
+    /// Conservative client-side policy for transport failures where a retry may succeed.
+    private static func isRetryableForwardPaymentTransportFailure(error: Error, statusCode: Int?) -> Bool {
+        if let code = statusCode {
+            if (500...599).contains(code) { return true }
+            if code == 408 || code == 429 { return true }
+            return false
+        }
+        let nsError = error as NSError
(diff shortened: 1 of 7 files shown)


thomson-t (dismissed):

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -682,6 +682,48 @@ class RoktInternalImplementation {
         return (false, failureReason)
     }
 
+    /// Conservative client-side policy for when the buyer may retry built-in card forward-payment
+    /// without leaving the confirm step (HTTP 200 with ``PurchaseResponse/success`` == `false`).
+    private static func isRetryableForwardPaymentBusinessFailure(failureReason: String?) -> Bool {
```
jamesnrokt: Given the terminology used elsewhere can we name the function to use the term card forwarding or similar?

```suggestion
    private static func isCardForwardingErrorRetryable(failureReason: String?) -> Bool {
```

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -682,6 +682,48 @@ class RoktInternalImplementation {
...
+            "service unavailable",
+            "too many requests",
+            "throttle",
+            "unavailable"
+        ]
+        return hints.contains { normalized.contains($0) }
+    }
+
+    /// Conservative client-side policy for transport failures where a retry may succeed.
+    private static func isRetryableForwardPaymentTransportFailure(error: Error, statusCode: Int?) -> Bool {
```
jamesnrokt: Same as above. Would be good to use the term card forwarding explicitly. May also make sense to extract some of these functions to a dedicated class as RoktInternalImplementation.swift is pretty chunky already

jamesnrokt (commented):
Overall looks good. Just wondering if we can break apart some of the longer functions here for simplicity

thomson-t (approved):

PR #196, opened 2026-06-12
refactor: convert TxnSessionManager to an actor

khushi1033:
## Background

`TxnSessionManager` holds the v2 init session state (session id, bearer token, expiry) and is read/written from concurrent async contexts. It previously guarded that state with a manual `NSLock`. This PR converts it to a Swift `actor` so the compiler enforces serialized access, removing the hand-rolled locking and the class of bugs that come with it.

Stacked on top of the v2-init wiring PR.

Fixes [(issue)]

## What Has Changed

- Converted `TxnSessionManager` from a `final class` + `NSLock` to an `internal actor`.
- Removed all manual `lock()/unlock()` calls; mutable state is now actor-isolated.
- `boundTagId` is `nonisolated let` so it stays synchronously readable.
- Renamed the private `isExpiredLocked` helper to `hasExpired` (locking no longer relevant).
- `TxnInitService` now `await`s the actor-isolated `authorizationHeader` and `update(...)`.
- Updated `TxnSessionManager` / `TxnInitService` unit tests to async.

## How Has This Been Tested

- Updated and ran the `TxnSessionManager` and `TxnInitService` unit test suites.
- Verified the v2 init flow still mints/persists a session and that the bearer  token round-trips through the actor.

## Notes

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
+++ b/Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
@@ -1,7 +1,7 @@
 // periphery:ignore:all - v2 session store; token-refresh/session-id members are consumed by the upcoming v2 offers/events path
 import Foundation
 
-internal final class TxnSessionManager {
+internal actor TxnSessionManager {
     // Persistence keys for the v2 path, separate from the legacy SessionManager's `rokt.*` keys.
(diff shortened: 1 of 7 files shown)


denischilik (approved):
LGTM

PR #201, opened 2026-06-16
feat: v2 cross-integration session sharing via token bundle

alexs-mparticle:
## Summary

Ports native ⇄ WebView **session sharing** into the v2 Transactions path. A session is shared as a **bearer-token bundle** `{sessionId, token, expiresAt}` (a "Shared Session"), **not** a bare session id — because `/v2/sessions/init` has no `session_id` field; continuity rides `Authorization: Bearer <JWT>` only. A receiving integration can therefore continue the *same authenticated session* on the Transactions Gateway.

> **Draft / iOS half only.** True end-to-end interop is **blocked on the WSDK token bridge** (see ADR + below). This PR is the iOS side, unit-tested in isolation and manually verified for the local seed→init→export round-trip.

## What's here

- **Models** — public `RoktSharedSession` (`@objc`) + internal `TxnSharedSession` (Equatable). `token` is a bearer credential: do not log/persist outside the SDK.
- **Public API** — `Rokt.setSharedSession(_:)` / `getSharedSession()`, alongside the legacy id-only `setSessionId`/`getSessionId` (v1 surface, unchanged).
- **`TxnSessionManager`** — non-mutating `sharedSession` export (nil if no live/expired session) + `seed(sharedSession:)` that adopts a bundle **iff unexpired** (server-authoritative expiry; never clobbers a live session with an expired seed).
- **`RoktInternalImplementation`** — holds a **pending seed** before init, then captures `service.sessionManager` at init and consumes the seed so its bearer rides `/v2/sessions/init`. Expired seed → ignored → init mints a fresh session.
- **ADR-0001** — records the token-bundle decision, backend contract evidence, and the WSDK gap.

## Design decisions

- **Token bundle, not a bare id.** Verified against both backends: the gateway can't continue a session from a client-supplied id; only a valid bearer continues it.
- **Import = a pre-init seed consumed *by* init**, not an alternative to it. iOS always calls init (also returns fonts/flags/timeout).
- **Server-authoritative expiry.** Adopt the token's expiry verbatim; drop an already-expired bundle. No legacy 30-min/50-call lifetime.

## Testing

- **Unit (green):** 7 seed/export cases in `TestTxnSessionManager` + 2 init-seeding cases in `TestTxnInitWiring` (live seed adopted by init's manager and bearer set; expired seed dropped → init mints `sess-1`). Run via the `Rokt-Widget` package scheme.
- **Manual (sim, Stage):** fresh-process round-trip confirmed the pending-seed path live — `BEFORE getSharedSession()=nil` → init captures the manager → `AFTER` exports the seeded session through production code. (Init returns failure only because the *synthetic* seed JWT is rejected by Stage — expected; a real token can't be minted client-side, hence the WSDK dependency.)
- **Manual (sim, Mock env, via Example app):** Wired a temporary hook into the Shoppable Ads demo's init flow and drove a full `setSharedSession` → `initWith` → `getSharedSession` round-trip through production code. The event log confirmed all three states end-to-end: the pending seed is surfaced before init, the server-issued token replaces it after a successful init, and a post-init `setSharedSession` overrides the live session (the token subsequent offers calls would use).

## Blocker / follow-up (not in this PR)

End-to-end interop needs parallel **WSDK** work: `sessionToken`+`expiresAt` on its init/inject path, a `SessionSource.NATIVE_INJECTED`, and a `getCurrentSession()` getter. Needs socializing with the web team.


docs/adr/0001-v2-session-sharing-carries-a-token-bundle.md
```diff
@@ -0,0 +1,29 @@
...
+- **WSDK** (`rokt-wsdk`): likewise authenticates subsequent calls by bearer token and keys stored tokens by id. A bare id continues nothing.
+
+In v2, the **Session Token (JWT) is the load-bearing credential** for continuity; the id is only for correlation and the WSDK's token-keying.
+
+## Decision
+
+1. **Share a token bundle, not an id.** v2 session sharing carries a **Shared Session** = `{ sessionId, token, expiresAt }` via new public methods `setSharedSession(RoktSharedSession)` / `getSharedSession() -> RoktSharedSession?`, added _alongside_ the legacy `setSessionId`/`getSessionId` (which remain as the v1 id-only surface). A bare `String` cannot carry the credential, so the existing signatures are not reused. The `expiresAt` lets the receiver check validity without decoding the token, which iOS treats as opaque.
```
thomson-t: Why do we need a `sessionId` passed separately when token has that exact same thing in the sub?
How can we trust they both match without verifying them?
Why do we need it `alongside` the `setSessionId`?
SDK is not going to keep the /experiences api in the new version and I don't think the /offers api can understand the legacy sessionId either.

docs/adr/0001-v2-session-sharing-carries-a-token-bundle.md
```diff
@@ -0,0 +1,29 @@
...
+
+2. **Import is a pre-init seed consumed by `/v2/sessions/init`, not an alternative to it.** iOS always calls init regardless of any shared session (init also returns fonts, feature flags, and timeout). `setSharedSession` seeds the token into the `TxnSessionManager` that init will use, _before_ init runs; the gateway then continues that session. An expired seed contributes no bearer (`authorizationHeader` is nil), so init mints a fresh session — i.e. an expired shared session is silently ignored. Export (`getSharedSession`) is a non-mutating read of the live manager. This removes any before/after-init ordering hazard: there is no race, only an input to the always-present init call.
```
thomson-t: I don't think `getSharedSession`/`getSessionId` is used by web sdk or any other consumers at the moment. But thats ok.
`setSharedSession` can happen after the init as well. This will override the session set by the init and the new session token from setSharedSession will be used in the offers api. Just wanted add this in the docs as an expected use case.

docs/adr/0001-v2-session-sharing-carries-a-token-bundle.md
```diff
@@ -0,0 +1,29 @@
...
+
+## Consequences
+
+- Two session-sharing surfaces coexist on the public API (legacy id-only + v2 Shared Session); doc-comments must steer v2 integrators to `setSharedSession`.
+- The host app remains responsible for moving the Shared Session across the native↔WebView boundary (no built-in bridge, unchanged from legacy).
+- The WSDK currently has **no** token import/export API (only a web↔web cookie/localStorage recogniser bridge) — matching web-side work is required for end-to-end interop and is tracked separately.
```
thomson-t: Should we add a TODO item to update this when it is done?

docs/adr/0001-v2-session-sharing-carries-a-token-bundle.md
alexs-mparticle: Good calls — addressed in `7e3bb6e` and `7c5d665`.

- **Dropped `sessionId` from the bundle.** You're right that the id already lives in the token's JWT `sub`, so carrying a separate caller-supplied id was redundant and unverifiable. `RoktSharedSession` is now token-only: `{ token, expiresAt }`. The SDK's internal session id is still sourced from the gateway init response — only the *shared bundle* loses it.
- **Removed the legacy id-only API entirely.** Agreed it's dead weight in v2 (`/experiences` going away, `/offers` can't continue from a client-supplied id). `Rokt.setSessionId`/`getSessionId` + impls + their tests are gone. Note this is a **breaking public-API change** (committed as `feat!:`) — flagging for release notes.

docs/adr/0001-v2-session-sharing-carries-a-token-bundle.md
alexs-mparticle: Documented in the ADR (`ec1e929`). Confirmed `setSharedSession` is supported **after** init as well — it overrides the session established by init, and the new token is what's used on the offers API. The implementation already backs this path (post-init adopt + re-init carry-forward), and there are now tests covering it.

docs/adr/0001-v2-session-sharing-carries-a-token-bundle.md
alexs-mparticle: Added a TODO in the consequences section (`ec1e929`) to update this ADR when the WSDK token-bridge work lands.

PR #202, opened 2026-06-17
feat: add v2 events enablement

khushi1033:
## Background

Adds a v2 events path that sends native SDK events to `POST /v2/sessions/events`, mirroring the existing v2 init wiring. Events are mapped to the v2 wire shape, the session JWT is attached when valid (and rotated from the response), and live sends are routed behind a toggle so the legacy events path stays available for rollback.

## What Has Changed

- New `TxnEventService` — sends a mapped batch, attaches `Authorization` when the token is valid (omits it when missing/expired so the server mints a session), rotates the returned `session_token`, retries on transient 5xx/transport errors, and soft-fails so events never block or crash.
- New `TxnEventMapper` — maps SDK event types to the v2 `event_type` vocabulary and builds the `data` map (attributes/metadata, `capture_method`, epoch-ms timestamp). Reserved keys (`parent_id`/`token`/`page_instance_guid`) are written last so partner attributes can't override them; `capture_attributes` nests partner data under `data.attributes`.
- `V2EventsClient` — `Authorization` is now optional, stamps the `rokt-*` device header subset, and `data` supports nested values for `capture_attributes`.
- `RoktInternalImplementation` — adds the `useTxnEvents` toggle plus the event service factory/test seam; `PlatformEventProcessor` and the `CaptureAttributes` callback branch their send seams on the toggle.

## How Has This Been Tested

- Unit tests for the mapper (event-type vocabulary, reserved-key precedence, nested capture attributes, timestamp), the service (auth attached/omitted, device headers, token rotation, retry/soft-fail), and the toggle wiring.
- Existing `/v2/sessions/events` consumer pact still validates the wire contract.
- Verified end-to-end in the example app — events send to the correct `/v2/sessions/events` endpoint with the expected headers and payloads.

## Notes
- Creating a followup PR to remove references to V2

## Checklist


Sources/Rokt_Widget/Networking/TxnEventMapper.swift
```diff
@@ -0,0 +1,131 @@
+import Foundation
+internal import RoktUXHelper
+
+// Maps domain events to the v2 wire shape. Reserved keys (parent_id, token, page_instance_guid)
+// are written last so partner attributes cannot overwrite them.
+internal enum TxnEventMapper {
```
thomson-t, line 6: Why do we need this mapper?
The UXHelper should be sending us the right events and the job of the SDK is just to send it to the backend API.

khushi1033: UXHelper doesn't emit events in the shape TxN expects. Let me know if we'd prefer to move the mapping to UXHelper, however, I believe, mapping in the SDK mirrors whats done on [android](https://github.com/ROKT/sdk-android-source/blob/af604f03d9362b4a1feaa27bd1d802fae0c44e6a/roktsdk/src/main/java/com/rokt/data/impl/repository/txn/EventMapper.kt).

Tests/Rokt_WidgetTests/TestTxnEventWiring.swift
```diff
@@ -0,0 +1,128 @@
...
+                sdkVersion: "5.2.2",
+                sessionManager: TxnSessionManager(),
+                httpClient: stub!,
+                baseBackoff: 0,
+                sleep: { _ in }
+            )
+        }
+    }
+
+    private func waitUntil(_ condition: @escaping () -> Bool, timeout: TimeInterval = 2) {
```
thomson-t: Rather than having private wait functions in each test files, can we have this in a generic place so that other tests can reuse it?
There is a similar function here  Tests/Rokt_WidgetTests/Extension/XCTestCase+Extension.swift

Tests/Rokt_WidgetTests/Shared/Networking/TestTxnEventService.swift
```diff
@@ -0,0 +1,236 @@
...
+        } catch let error as TxnEventService.TxnEventError {
+            XCTAssertEqual(error, .invalidBaseURL)
+            XCTAssertEqual(httpClient.callCount, 0)
+        } catch {
+            XCTFail("Unexpected error: \(error)")
+        }
+    }
+}
+
+private final class MockTxnEventsHTTPClient: HTTPClientAdapter {
```
thomson-t: Maybe extract this as well and make a Tests/Rokt_WidgetTests/Shared/Networking/MockTxnEventsHTTPClient.swift so that we can reuse it instead of StubEventsHTTPClient

Sources/Rokt_Widget/Networking/TxnEventService.swift
```diff
@@ -0,0 +1,118 @@
...
+        httpClient.updateTimeout(timeout: requestTimeout)
+
+        let client = V2EventsClient(
+            baseURL: baseURL,
+            accountId: accountId,
+            authToken: sessionManager.authorizationHeader,
+            sdkVersion: sdkVersion,
+            deviceHeaders: deviceHeaders,
+            httpClient: httpClient
+        )
```
thomson-t: We don't need to do these operations every time when the send method is called. This can be done at the time of the instantiation.

Tests/Rokt_WidgetTests/TestTxnEventWiring.swift
khushi1033: Good call, moved it into `XCTestCase+Extension.swift `and removed the copies from both test files

Tests/Rokt_WidgetTests/Shared/Networking/TestTxnEventService.swift
khushi1033: Done! Extracted it to `Tests/Rokt_WidgetTests/Shared/Networking/MockTxnEventsHTTPClient.swift` and switched the wiring test over to it

Sources/Rokt_Widget/Networking/TxnEventService.swift
khushi1033: Done! Moved the URL parsing, timeout, and client creation into init

thomson-t (approved):

PR #205, opened 2026-06-17
fix: keep in-app browser open for internal links on the last offer

nickolas-dimitrakas:
## Background

When a positive-CTA link is configured to open **internally** (in-app `SFSafariViewController`), tapping it on the **last/only offer** caused the in-app browser to appear and then immediately dismiss itself within about a second. Earlier offers were unaffected.

Root cause: the in-app browser is presented on top of the placement overlay (the overlay is its presenter). The link completion — which advances/closes the offer — was fired as soon as Safari finished presenting. On the last/only offer that completion closes the placement by calling `dismiss()` on the overlay. Because `dismiss()` dismisses whatever a controller is currently presenting, and the overlay was presenting Safari, UIKit dismissed **Safari** instead of the overlay. On non-last offers the completion only transitions to the next offer (no `dismiss()`), which is why the issue was specific to the last/only offer.

Fixes [(issue)]

## What Has Changed

- `LinkHandler` now becomes the `SFSafariViewControllerDelegate` and defers the link completion until the in-app browser is actually dismissed by the user (`safariViewControllerDidFinish`), instead of firing it the moment Safari finishes presenting.
- `LinkHandler` now inherits from `NSObject` (required to conform to `SFSafariViewControllerDelegate`).
- Added a regression test asserting that, for an internally-opened link, Safari stays presented and the completion only fires after the browser is dismissed.

## How Has This Been Tested

- Added/ran a unit test (`ValidLayoutOverlayTests` → "internal link defers completion until Safari is dismissed") on the iOS Simulator that drives the real `LinkHandler` against a presented placement overlay and verifies the deferred-completion behaviour.
- Manually traced and verified on the simulator that the in-app browser now stays open on the last/only offer and the placement closes cleanly only after the browser is dismissed.

## Notes

Behaviour change is limited to internally-opened links: the completion now runs on browser dismissal rather than on browser presentation, which also matches the intent of the close callback. External/passthrough link handling is unchanged.

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/Util/LinkHandler.swift
+++ b/Sources/Rokt_Widget/Util/LinkHandler.swift
@@ -2,12 +2,14 @@ import Foundation
 import SafariServices
 internal import RoktUXHelper
 
-class LinkHandler {
+class LinkHandler: NSObject {
     private static let urlDiagnosticCode = "[URL]"
 
     private var completionHandler: (() -> Void)?
 
-    init() {}
+    override init() {
+        super.init()
+    }
 
     private func openURL(url: URL, type: RoktUXOpenURLType) {
         func openExternalLink(_ url: URL) {
@@ -29,9 +31,8 @@ class LinkHandler {
             }
             let safariVC = SFSafariViewController(url: url)
             safariVC.modalPresentationStyle = .overFullScreen
-            UIApplication.topViewController()?.present(safariVC, animated: true, completion: { [weak self] in
-                self?.completionHandler?()
-            })
+            safariVC.delegate = self
+            UIApplication.topViewController()?.present(safariVC, animated: true)
         case .externally,
                 .passthrough:
             completionHandler?()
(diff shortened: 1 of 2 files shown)


Sources/Rokt_Widget/Util/LinkHandler.swift
```diff
@@ -60,3 +61,15 @@ class LinkHandler {
     }
 
 }
+
+extension LinkHandler: SFSafariViewControllerDelegate {
```
BrandonStalnaker, line 65: Is there a reason you put this in an extension rather than just the class?

nickolas-dimitrakas: No functional reason — it's a stylistic choice to keep the `SFSafariViewControllerDelegate` conformance grouped in its own extension along with the explanatory comment, which keeps the core class body focused on the link-opening logic. It also matches the existing convention in the codebase (e.g. `RoktEmbeddedView: InternalLayoutLoader` and `RoktLayoutViewModel: LayoutLoader` are both done this way). Happy to fold it into the class body if you'd prefer.

thomson-t (approved):

PR #208, opened 2026-06-19
feat(offers): fetch, adapt, render, and wire v2 offers end to end

alexs-mparticle:
## Background

The iOS v2 offers path, end to end: fetch offers from `POST /v2/sessions/offers`, adapt the response into the experience the renderer understands, and route it through the live `execute` path. It consolidates the stacked offers work (#212, #213, #214, #215) on top of `workstation/transactions`, inheriting that base's v2 init/events work and the `TxnSessionManager` actor.

## What has changed

**Wire models (`SelectResponse.swift`)**
- Request/response models for the offers endpoint. Session identity comes from the JWT `sub` claim in the `Authorization` header, not a body `session_id`; privacy travels under `privacy_control`, and `customer` / `page.url` are omitted to match Android.
- DCUI schemas (`outer_layout_schema`, `layout_variant_schema`) stay as pre-serialized strings, the shape the renderer expects. Real-time events and `catalog_items` are typed rather than opaque JSON. Decoding stays lenient because the provider emits `omitempty` throughout.

**Request shape**
- `OffersClient` / `OffersService` build and send the request with retry/backoff mirroring init, read the JWT from a rehydrated `TxnSessionManager`, and roll the refreshed `session_token` forward.
- The request carries what server-side page detection and targeting need: `rokt-package-name` (the bundle id; detection 409s without it), `channel.rokt_platform_type` (`iOS`), and the full mobile device header set via `NetworkingHelper.txnDeviceHeaders()` (`rokt-os-type`, `rokt-os-version`, `rokt-device-model`, `rokt-ui-locale`, `rokt-package-version`).
- `privacy.gpc_enabled` rides as a top-level sibling of `privacy_control`.
- Real-time `events[]` triggered on the prior placement are forwarded, and the response's `event_data` is captured for the next call.
- `Authorization` is omitted when there is no live token, so the server mints a session.
- `MockOffersHTTPClient` serves an offline fixture so Mock builds still exercise the real decode, adapt, and render path.

**Render adapter (`SelectExperienceAdapter`)**
- Maps the offers response into the experience JSON the renderer decodes: re-homes keys to camelCase, passes DCUI schemas through verbatim, buckets response options positive/negative, orders keys stably, and drops an offer when its creative is absent.

**Wiring (`RoktInternalImplementation`)**
- The offers call site routes through `OffersService` and hands the adapted experience to the existing render path, recording `pageInit` timing at the call site so timing parity with v1 holds (the service does not extract timing itself). The tracking-consent gate is preserved, and a `useV2Offers` seam (default v2) keeps the v1 fetch a live alternative rather than dead code.

**Shared plumbing and hardening**
- `TxnRequestHeaders` is the single owner of the headers common to every v2 request (account id, content type, optional `Authorization`), so init/events/offers cannot drift.
- `RealtimeEventStore` writes are de-duped, bounded, and serialized on the event queue, so echoed `event_data` cannot grow the store unbounded or be lost under concurrent responses. The epoch-ms formatting and the event flatten/validate rule are each shared between the v1 and v2 paths.

**Contracts and CI**
- The consumer can-i-deploy gate is now blocking against production (previously an advisory no-op pointed at a `stage` placeholder), so a non-deployable contract fails the PR; this matches Android.
- Contract content is brought to parity: a second offers interaction for the no-page-detection state, cold-init modeling, exact feature-flag matching, a UUID-matched events `instance_id`, and `is_page_detected` pinned to exact `true`.
- The pact path filter now triggers on `OffersClient` and the `SelectResponse` wire model (replacing the removed `TxnOffersClient`), with a `pull-requests: read` scope so it can list changed files. PR tests target the iPhone 17 Pro simulator, since the runner image dropped iPhone 16 Pro.

**Naming**
- Offers networking carries no version markers (`OffersClient` / `OffersService` / `SelectResponse`), replacing the prior `TxnOffersClient` / `TxnOffersClientPactSpec` (removed, hence the large deletion in the diff), so it reads as the canonical path alongside the base's `Txn`-prefixed init/events.

## How this has been tested

- Full unit suite green on the iOS simulator (579 tests), including the offers pact (both interactions), a render-compat test (adapter output parses to a non-nil `RoktUX` page model with real DCUI schemas), and end-to-end `execute` wiring tests (success, failure, Mock transport, cache write then reuse).
- The existing Mock cache/execute suite (`RoktExperienceCacheExecuteTests`) was migrated to assert against the v2-adapted page, so it now drives the full reshape, decode, adapt, render path, backed by a `makeOffersData(fromV1Experience:)` fixture reshaper and an `expectedOffersPage` helper.
- Verified against the live provider contract: `/v2/sessions/offers` returns 200 with session and token roll-forward, carrying the package name, platform, and device signals page detection needs.

Tested locally
Screenshot 2026-06-26 at 8 39 31 AM

## Notes

- The full `SelectCatalogItem` wire model (pricing, quantity, images, tokens) is decoded; only its mapping into the render models is deferred to the mapper.
- `rokt-txn-shadow: false` is no longer sent on v2 requests — mobile shadow routing is session-id-shape driven, not header-driven. `/v2/sessions/events` now sends `single_session: true` (one session per batch, matching the web controller), and the events pact asserts it.
- Only `rokt-package-name` of the device headers is asserted in the published pact. The other device headers, `events[]`, and `gpc_enabled` ride the live request but are not asserted yet, since asserting them changes the contract and needs provider re-verification first.

## Checklist


Tests/Rokt_WidgetTests/Resource/offers.json
thomson-t, line 1: Is this used somewhere?

Example/Tests/txn_offers.json
```diff
@@ -0,0 +1,77 @@
...
+                      "instance_guid": "mock-response-option-instance-1",
+                      "token": "mock-response-option-token",
+                      "signal_type": "signal-response",
+                      "short_label": "Yes please",
+                      "is_positive": true,
+                      "url": "https://example.com/accept"
+                    },
+                    "negative": {
+                      "id": "mock-response-option-negative",
+                      "action": "noFunc",
```
thomson-t: Is this a valid action?

Sources/Rokt_Widget/Models/SelectResponse.swift
```diff
@@ -0,0 +1,347 @@
...
+        outerLayoutSchema = try container.decodeIfPresent(String.self, forKey: .outerLayoutSchema)
+        token = try container.decodeIfPresent(String.self, forKey: .token)
+    }
+}
+
+internal struct TxnSelectSlot: Decodable, Equatable {
+    let instanceGuid: String?
+    let layoutVariant: TxnSelectLayoutVariant?
+    let offer: TxnSelectOffer?
+    let token: String?
```
thomson-t, line 208: Not sure if we need to tolerate nullable here. Why would the API send these as null when there is a slot? Can you check the API code and make them non-null if they can't be null?

Sources/Rokt_Widget/Models/SelectResponse.swift
```diff
@@ -0,0 +1,347 @@
...
+    let action: String?
+    let instanceGuid: String?
+    let token: String?
+    let signalType: String?
+    let shortLabel: String?
+    let longLabel: String?
+    let shortSuccessLabel: String?
+    let isPositive: Bool
+    let url: String?
+    let ignoreBranch: Bool?
```
thomson-t, line 328: Any particular reason why these fields are nullable? 
If something we expect are null eg: token, we should fail early and stop there, rather than propagating it to other layers.

Sources/Rokt_Widget/Models/TxnSelectResponse.swift
```diff
@@ -0,0 +1,347 @@
...
+}
+
+internal struct TxnSelectIcon: Decodable, Equatable {
+    let name: String?
+}
+
+/// Token lookup for a trackable entity, echoed back on `/v2/sessions/events`.
+internal struct TxnSelectEventDataEntry: Decodable, Equatable {
+    let token: String
+    let events: [String: TxnJSONValue]?
```
thomson-t: Is this an unknown type?

Tests/Rokt_WidgetTests/Resource/offers.json
crisryantan: Yes — it's the offline fixture the Mock-environment transport serves, and it's decode-guarded by `TestOffersFixture`.
https://github.com/ROKT/rokt-sdk-ios/blob/a21d383/Sources/Rokt_Widget/Networking/MockOffersHTTPClient.swift#L84-L86

Example/Tests/txn_offers.json
crisryantan: The invalid values are gone. The fixture (now `offers.json`) uses renderer-recognized actions (`Url`/`CaptureOnly`), and `TestOffersFixture` now fails if any action or signal_type isn't one the renderer maps.
https://github.com/ROKT/rokt-sdk-ios/blob/a21d383/Tests/Rokt_WidgetTests/Shared/Model/TestOffersFixture.swift#L48-L62

Sources/Rokt_Widget/Models/SelectResponse.swift
crisryantan: Checked the provider — it emits `omitempty` across the offers response, so a present slot can still legitimately omit these. We kept the decode lenient (documented on `SelectResponse`) and made `slots` itself non-optional/defaulted; the adapter drops structurally-incomplete elements (e.g. an offer with no creative) rather than the decoder rejecting otherwise-valid payloads.
https://github.com/ROKT/rokt-sdk-ios/blob/a21d383/Sources/Rokt_Widget/Models/SelectResponse.swift#L96-L105

Sources/Rokt_Widget/Models/SelectResponse.swift
crisryantan: Same reason on the provider side (`omitempty` throughout). The adapter is the gate rather than the decoder: it drops an offer with no creative and a response-option map with neither a positive nor negative option, so partials never reach the render layer. `isPositive` is the one field we always need, so it's decoded with a default.
https://github.com/ROKT/rokt-sdk-ios/blob/a21d383/Sources/Rokt_Widget/Networking/SelectExperienceAdapter.swift#L135-L138

Sources/Rokt_Widget/Models/TxnSelectResponse.swift
crisryantan: Replaced — the opaque JSON value is now a typed `SelectRealTimeEvent` (`event_type` + pre-serialized `payload`), decode-guarded by `TestOffersFixture`.
https://github.com/ROKT/rokt-sdk-ios/blob/a21d383/Sources/Rokt_Widget/Models/SelectResponse.swift#L366-L376

Sources/Rokt_Widget/Networking/OffersClient.swift
```diff
@@ -0,0 +1,113 @@
...
+
+internal struct OffersInput {
+    let requestId: String
+    let pageIdentifier: String
+    let attributes: [String: String]
+    let privacyControl: SelectPrivacyControl?
+    let privacy: SelectPrivacy?
+    let events: [SelectEvent]?
+
+    init(
```
denischilik: Looks like we don't need constructor here swift should generate it itself

Sources/Rokt_Widget/Networking/OffersService.swift
```diff
@@ -0,0 +1,231 @@
...
+    ) {
+        onRequestStart?()
+
+        // Extract privacy KVPs before sanitising, then enrich the remaining attributes.
+        // gpc_enabled travels under `privacy`, separate from `privacy_control`, to match Android.
+        let privacyPayload = RoktAPIHelper.getPrivacyControlPayload(attributes: attributes)
+        let privacyControl = buildPrivacyControl(from: privacyPayload)
+        let privacy = buildPrivacy(from: privacyPayload)
+        let sanitisedAttributes = RoktAPIHelper.removePrivacyControlAttributes(attributes: attributes)
+        let enrichedAttributes = AttributeEnrichment.shared.enrich(attributes: sanitisedAttributes, config: config)
```
denischilik: Consider to inject AttributeEnrichment.shared instead have it hardcoded in the class

Sources/Rokt_Widget/Networking/SelectExperienceAdapter.swift
```diff
@@ -0,0 +1,234 @@
...
+        guard let string = String(data: data, encoding: .utf8) else {
+            throw AdapterError.encodingFailed
+        }
+        return string
+    }
+}
+
+// MARK: - Renderer experience contract (camelCase)
+
+private struct RenderExperience: Encodable {
```
denischilik: Render* is the renderer contract (RoktUXExperienceResponse), not networking — it belongs in Models/, alongside SelectResponse. Extract it now while the mapping is new; moving it later will be a noisier refactor.

Sources/Rokt_Widget/Networking/SelectExperienceAdapter.swift
```diff
@@ -0,0 +1,234 @@
...
+private struct RenderPage: Encodable {
+    let pageId: String?
+}
+
+private struct RenderPlacementContext: Encodable {
+    let roktTagId: String
+    let pageInstanceGuid: String
+    let token: String
+
+    init(_ r: SelectResponse) {
```
denischilik: Single-letter params (r, p, c…) aren't used elsewhere in the SDK — mapping inits typically use full names (event, address, contactAddress). Consider response / plugin / config for consistency.

.github/workflows/pull-request.yml
```diff
@@ -92,7 +92,7 @@ jobs:
         run: |
           xcodebuild test \
             -scheme Rokt-Widget \
-            -destination 'platform=iOS Simulator,name=iPhone 16 Pro,OS=latest' \
```
denischilik, line 95: Looks like updating simulator is not related to this PR

denischilik (commented):
One more comment

Sources/Rokt_Widget/Networking/OffersClient.swift
crisryantan: Done in 011ec52 — dropped the explicit init on `OffersClient` and `OffersInput` and let Swift synthesize it. Carried the same across the other new offers structs (`SelectRequest`, `SelectChannel`, `SelectPrivacyControl`, `SelectPrivacy`). Where a default was relied on at a call site, the field is now `var` so the synthesized memberwise init keeps the same call sites.

Sources/Rokt_Widget/Networking/OffersService.swift
crisryantan: Done in 011ec52 — `AttributeEnrichment` is now an injected dependency on `OffersService` (defaults to `.shared`), matching the existing `httpClient` / `sessionManager` seams.

Sources/Rokt_Widget/Networking/SelectExperienceAdapter.swift
crisryantan: Done in 011ec52 — moved the `Render*` types into `Models/RenderExperience.swift`, next to `SelectResponse`. The adapter just performs the encode now.

Sources/Rokt_Widget/Networking/SelectExperienceAdapter.swift
crisryantan: Done in 011ec52 — renamed to full names (`response`, `plugin`, `config`, `slot`, `offer`, `creative`, ...). Applied the same to the offers mock reshapers in `MockResponse.swift` for consistency.

.github/workflows/pull-request.yml
crisryantan: Good catch — reverted to iPhone 16 Pro (matches the base branch) in 011ec52.

crisryantan: Update — keeping iPhone 17 Pro after all. The CI runner image moved to Xcode 26, which ships no iPhone 16 Pro, so reverting red the Package SPM Tests (`Unable to find a device matching ... iPhone 16 Pro`). The bump is required for the current runner, not unrelated cruft — restored in 7b32dda.

denischilik (approved):

PR #214, opened 2026-06-26
pact: bring iOS consumer to parity with Android (blocking gate + contract content)

crisryantan:
## What

Bring the iOS Pact Consumer flow **and** contract content to parity with the Android consumer (`sdk-android-source`), and flip `can-i-deploy` from advisory to **blocking**.

### Commit 1 — flow parity + blocking gate (`ci(pact)`)
- `can-i-deploy` now **fails the PR** when the published contract isn't verified deployable to `production` (was advisory against a `stage` placeholder that doesn't exist in the broker, so it always took the warning branch). `set -euo pipefail`, dropped the warning-swallow wrapper, job renamed `Can I Deploy`.
- Fix stale path filter `TxnOffersClient.swift` → `OffersClient.swift`; watch the offers request models (`SelectResponse.swift`); add `pull-requests: read`; bump `actions/checkout` to v6.0.3.

This commit doesn't change the published pact, so it verified **green** (`Computer says yes \o/`) — confirming the blocking gate works end-to-end.

### Commit 2 — contract-content parity (`test(pact)`)
- **offers**: add the `no page detection` interaction (session + token, `is_page_detected:false`, no plugins).
- **init**: model a cold init — drop the `Authorization` assertion and invoke the client without a token (the server mints the session); keep `x-request-id`, which the iOS client always sends.
- **init**: exact-match the four feature flags the provider returns verbatim.
- **events**: assert `instance_id` is a canonical UUID (regex) rather than any string.

Each change was validated against the `transactions` provider stub and verifies green.

## Why `Can I Deploy` is red right now (expected)

Commit 2 changes the published pact, so the gate reports:
> There is no verified pact between this version of rokt-sdk-ios and the version of transactions-api currently in production.

That's the blocking gate doing its job — the new contract hasn't been provider-verified yet. **No provider code change is needed** (all changes verify green against the current stub). It recovers to green once a `transactions` **master** build runs `pact:verify` (verifies the new contract — no branch filter, runs on every build) and `pact:record-deployment` (records the prod deployment — master/hotfix only).

## Notes
- Stacks on #208 (base `feat/v2-offers-network`). Draft until the gate recovers green after provider re-verification.
- Once reliably green, add `Can I Deploy` to required status checks (branch protection) so it blocks merges.

### DIFF
--- a/.github/workflows/pact-consumer.yml
+++ b/.github/workflows/pact-consumer.yml
@@ -1,5 +1,8 @@
 name: Pact Consumer
 
+# Generates the consumer pact on every PR, publishes it to the broker on
+# same-repo PR pushes and main, then runs a blocking can-i-deploy check.
+
 on:
   pull_request:
   push:
@@ -7,6 +10,7 @@ on:
 
 permissions:
   contents: read
+  pull-requests: read # paths-filter lists changed files via the PR files API
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -20,7 +24,7 @@ jobs:
       contract: ${{ steps.filter.outputs.contract }}
     steps:
       - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           persist-credentials: false
           # paths-filter needs history to diff against the previous commit on push.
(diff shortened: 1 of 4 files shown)


.github/workflows/pact-consumer.yml
```diff
@@ -32,9 +36,10 @@ jobs:
         with:
           filters: |
             contract:
-              - 'Sources/Rokt_Widget/Networking/TxnOffersClient.swift'
+              - 'Sources/Rokt_Widget/Networking/OffersClient.swift'
```
alexs-mparticle, line 39: Question: I thought we needed to keep `Txn` prefix for iOS? Not a blocker, just legit curious.

crisryantan: I think the agreement on android side was different. afaik we have to get rid of v2/txn prefixes/suffixes since offers would be the source of truth once `workstation/transactions` gets merged. since I was already working on it, I went with parity on the android implementation

crisryantan: probably a conflicting thing was that it was also agreed clean-up should be done later, but meh, it was a quick change to do anyway so I did it now

alexs-mparticle (approved):

PR #215, opened 2026-06-26
feat(offers): mobile device headers and privacy.gpc_enabled on the v2 offers request

crisryantan:
## Summary

Fills the remaining gaps in the v2 offers request shape so it matches Android: the mobile device headers and the GPC consent signal now ride on `POST /v2/sessions/offers`.

## Changes

**Device headers.** `deviceHeaders` is plumbed through `OffersService` → `OffersClient`, and `defaultOffersService` passes `NetworkingHelper.txnDeviceHeaders()` just like `defaultTxnEventService`. Adds `rokt-os-type`, `rokt-os-version`, `rokt-device-model`, `rokt-ui-locale`, `rokt-package-version`. `rokt-package-name` now rides in via the device headers (set only when the bundle id resolves) instead of a hardcoded `Bundle.main.bundleIdentifier ?? ""` with a silent empty fallback.

**Authorization.** Omitted entirely when there is no live session token (was sending a blank `Bearer`), matching the events client — the server then mints a fresh session.

**privacy.gpc_enabled.** The GPC signal was parsed from the attributes but dropped before the wire. It now travels as a top-level `privacy { gpc_enabled }` object, a sibling of `privacy_control` (not folded into it), omitted when absent. `privacy_control` is also omitted when none of its three keys are present (was encoding an empty object).

## Notes

The new device headers and `privacy.gpc_enabled` ride on the live request but are not asserted in the consumer pact yet, so the published contract stays at its current shape; the assertions are a follow-up. Real-time `events[]` forwarding is the remaining offers-request gap, handled in a separate stacked PR.

## Testing

`xcodebuild test` on the iPhone 17 Pro simulator — `TestOffersService`, `TestSelectSerialization`, `OffersClientPactSpec`, `TestOffersExecuteWiring`, `TestMockOffersHTTPClient` all green.

### DIFF
--- a/Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
+++ b/Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
@@ -15,7 +15,7 @@ class RealtimeEventStoreMemory: RealTimeEventStore {
     private var triggeredEvents: [TriggeredRealTimeEvent] = []
 
     func addUntriggeredEvents(_ events: [UntriggeredRealTimeEvent]) {
-        self.untriggeredEvents.append(contentsOf: events)
+        appendDeduped(events, to: &untriggeredEvents)
     }
 
     func getTriggeredEvents() -> [TriggeredRealTimeEvent] {
@@ -68,7 +68,7 @@ class RealTimeEventStoreFile: RealTimeEventStore {
     func addUntriggeredEvents(_ events: [UntriggeredRealTimeEvent]) {
         guard let untriggeredEventsFilePath else { return }
         var all = getUntriggeredEvents()
-        all.append(contentsOf: events)
+        appendDeduped(events, to: &all)
         save(all, to: untriggeredEventsFilePath)
     }
 
@@ -194,4 +194,14 @@ private func updateTriggeredEvents(
 private func trimTriggeredEvents(_ triggeredEvents: [TriggeredRealTimeEvent]) -> [TriggeredRealTimeEvent] {
     let sortedEvents = triggeredEvents.sorted { $0.eventTime > $1.eventTime }
     return Array(sortedEvents.prefix(maximumRealTimeEventsToStore))
+}
+
+// Appends only events not already present, preserving order. The backend can echo the same
+// event_data across responses; de-duping keeps the untriggered store bounded and stops a single
+// trigger from matching duplicate rows and multiplying what gets forwarded.
+private func appendDeduped(_ events: [UntriggeredRealTimeEvent], to all: inout [UntriggeredRealTimeEvent]) {
+    var seen = Set(all)
+    for event in events where seen.insert(event).inserted {
+        all.append(event)
+    }
 }
--- a/Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
+++ b/Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
@@ -16,6 +16,9 @@ class RealtimeEventStoreMemory: RealTimeEventStore {
 
     func addUntriggeredEvents(_ events: [UntriggeredRealTimeEvent]) {
         appendDeduped(events, to: &untriggeredEvents)
+        if untriggeredEvents.count > maximumRealTimeEventsToStore {
+            untriggeredEvents = Array(untriggeredEvents.suffix(maximumRealTimeEventsToStore))
+        }
     }
 
     func getTriggeredEvents() -> [TriggeredRealTimeEvent] {
@@ -67,9 +70,19 @@ class RealTimeEventStoreFile: RealTimeEventStore {
 
     func addUntriggeredEvents(_ events: [UntriggeredRealTimeEvent]) {
         guard let untriggeredEventsFilePath else { return }
-        var all = getUntriggeredEvents()
-        appendDeduped(events, to: &all)
-        save(all, to: untriggeredEventsFilePath)
+        // Serialize on the same queue as markAsTriggered's processing: this is a
+        // read-modify-write, so concurrent captures (or a capture racing a trigger-mark)
+        // would otherwise lose updates when the second save overwrites the first.
+        eventProcessingQueue.sync {
+            var all = getUntriggeredEvents()
+            appendDeduped(events, to: &all)
+            // Bound the untriggered file the way triggered events are capped: a long-lived
+            // session whose responses echo distinct event_data must not grow without limit.
+            if all.count > maximumRealTimeEventsToStore {
+                all = Array(all.suffix(maximumRealTimeEventsToStore))
+            }
+            save(all, to: untriggeredEventsFilePath)
+        }
     }
 
     func getTriggeredEvents() -> [TriggeredRealTimeEvent] {
--- a/Sources/Rokt_Widget/Networking/OffersClient.swift
+++ b/Sources/Rokt_Widget/Networking/OffersClient.swift
@@ -51,18 +51,9 @@ internal struct OffersClient {
             throw OffersClientError.bodyEncodingFailed
         }
 
-        var headers: RoktHTTPHeaders = [
-            "rokt-account-id": accountId,
-            "x-request-id": input.requestId,
-            "rokt-page-instance-guid": pageInstanceGuid,
-            // Live (non-shadow) v2 request.
-            "rokt-txn-shadow": "false",
-            "Content-Type": "application/json"
-        ]
-        // Authorization is optional: with no live token the server mints a fresh session.
-        if let authToken, !authToken.isEmpty {
-            headers["Authorization"] = authToken
-        }
+        var headers = TxnRequestHeaders.common(accountId: accountId, authToken: authToken)
+        headers["x-request-id"] = input.requestId
+        headers["rokt-page-instance-guid"] = pageInstanceGuid
         // Device headers (os, model, locale, app version) and the load-bearing
         // rokt-package-name ride in via NetworkingHelper.txnDeviceHeaders — the
         // gateway's mobile page detection and targeting key off them.
--- a/Sources/Rokt_Widget/Networking/OffersClient.swift
+++ b/Sources/Rokt_Widget/Networking/OffersClient.swift
@@ -42,7 +42,8 @@ internal struct OffersClient {
             page: SelectPage(pageIdentifier: input.pageIdentifier),
             channel: SelectChannel(sdkVersion: sdkVersion),
             attributes: input.attributes,
-            privacyControl: input.privacyControl
+            privacyControl: input.privacyControl,
+            privacy: input.privacy
         )
         let bodyData = try JSONEncoder().encode(requestBody)
         guard let bodyParameters = try JSONSerialization.jsonObject(with: bodyData) as? RoktHTTPParameters else {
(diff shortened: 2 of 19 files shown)


nickolas-dimitrakas (approved):
LGTM

Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
```diff
@@ -195,3 +195,13 @@ private func trimTriggeredEvents(_ triggeredEvents: [TriggeredRealTimeEvent]) ->
     let sortedEvents = triggeredEvents.sorted { $0.eventTime > $1.eventTime }
     return Array(sortedEvents.prefix(maximumRealTimeEventsToStore))
 }
+
+// Appends only events not already present, preserving order. The backend can echo the same
```
alexs-mparticle, line 212: The untriggered store isn't actually bounded — `appendDeduped` only suppresses exact duplicates; distinct events accumulate forever. Only triggered events are capped (`trimTriggeredEvents`, 50); `untriggered_events.json` is never trimmed. A long-lived session where the gateway echoes distinct `event_data` grows memory and disk without limit, and each file capture re-decodes/re-Sets/re-writes the whole array.

Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
```diff
@@ -68,7 +68,7 @@ class RealTimeEventStoreFile: RealTimeEventStore {
     func addUntriggeredEvents(_ events: [UntriggeredRealTimeEvent]) {
         guard let untriggeredEventsFilePath else { return }
         var all = getUntriggeredEvents()
-        all.append(contentsOf: events)
+        appendDeduped(events, to: &all)
```
alexs-mparticle: This capture path is an unsynchronized read-modify-write (`getUntriggeredEvents` → `appendDeduped` → `save(.atomic)`) run off `eventProcessingQueue`, unlike `markAsTriggered`. Two offers responses completing near-simultaneously each read, append to their own copy, and the second `save` overwrites the first — silently dropping captured events. Consider dispatching onto the existing serial queue.

Sources/Rokt_Widget/Networking/OffersService.swift
```diff
@@ -89,27 +108,38 @@ internal struct OffersService {
...
             pageInstanceGuid: makePageInstanceGuid(),
+            deviceHeaders: deviceHeaders,
             httpClient: httpClient
         )
+        // Forward events triggered during the previous placement; read once, before retries,
+        // and only with a live session to attribute them to (matching Android). As on the v1
+        // path, triggered events are not cleared after forwarding — they ride subsequent
+        // requests until session invalidation; re-send is expected (the "only adds, no clear"
+        // note elsewhere is about the untriggered response-capture, not this read).
+        let forwardedEvents = authToken != nil ? SelectEventMapper.requestEvents(from: triggeredEvents()) : []
```
alexs-mparticle, line 136: `triggeredEvents()` reads the entire persistent `RealTimeEventManager.shared` store with no session filter, and v2 advances the session via `sessionManager.update(sessionToken:)` (L166) — which never calls `sessionInvalidated()`, the only thing that clears the store. So events triggered in an earlier placement persist across launches and ride every subsequent offers request under whatever token is current. Worth confirming the gateway is idempotent on cross-session re-sends.

Sources/Rokt_Widget/Networking/SelectEventMapper.swift
```diff
@@ -0,0 +1,45 @@
...
+            )
+        }
+    }
+
+    /// Response `event_data` → untriggered events for the store, consulted when the
+    /// next placement fires. The flattening parallels ``UntriggeredEventsContainer`` (the
+    /// v1 path), but stays separate because v1 decodes camelCase `eventType` while v2
+    /// decodes snake_case `event_type`. Entries missing required fields are dropped here,
+    /// matching that sibling.
+    static func untriggeredEvents(from eventData: [String: SelectEventDataEntry]) -> [UntriggeredRealTimeEvent] {
```
alexs-mparticle, line 27: This duplicates the flatten/validate logic in `UntriggeredEventsContainer`. The camelCase-vs-snake_case difference justifies two Decodable types, not two copies of the flatten algorithm — a shared helper over already-decoded fields would avoid v1/v2 drift when the validity rule changes.

Sources/Rokt_Widget/Networking/OffersClient.swift
```diff
@@ -39,22 +42,33 @@ internal struct OffersClient {
...
-        let headers: RoktHTTPHeaders = [
+        var headers: RoktHTTPHeaders = [
             "rokt-account-id": accountId,
-            "Authorization": authToken,
             "x-request-id": input.requestId,
             "rokt-page-instance-guid": pageInstanceGuid,
-            // The app bundle id scopes server-side page detection for mobile.
-            "rokt-package-name": Bundle.main.bundleIdentifier ?? "",
+            // Live (non-shadow) v2 request.
+            "rokt-txn-shadow": "false",
```
alexs-mparticle: `rokt-txn-shadow: "false"` (and the optional-`Authorization` rule below) is pasted into three clients (here, TxnEventsClient, TxnInitClient) with no single owner. A partial edit when toggling/renaming shadow mode sends live traffic on one endpoint and shadow on another for the same session. A shared `txnCommonHeaders(...)` (sibling to v1's `getCommonHeaders`) is the right home.

Sources/Rokt_Widget/Networking/OffersService.swift
```diff
@@ -145,14 +179,23 @@ internal struct OffersService {
...
         return SelectPrivacyControl(
             noFunctional: payload[RoktAPIHelper.noFunctionalKey],
             noTargeting: payload[RoktAPIHelper.noTargetingKey],
             doNotShareOrSell: payload[RoktAPIHelper.doNotShareOrSellKey]
         )
     }
 
+    // gpc_enabled is a sibling of privacy_control (Android parity), omitted when the partner sent none.
+    private func buildPrivacy(from attributes: [String: String]) -> SelectPrivacy? {
+        let payload = RoktAPIHelper.getPrivacyControlPayload(attributes: attributes)
```
alexs-mparticle: `getPrivacyControlPayload(attributes:)` is computed twice per request — `buildPrivacyControl` (L181) and `buildPrivacy` here, called back-to-back. Compute the payload once and pass it into both.

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -972,7 +972,8 @@ class RoktInternalImplementation {
             accountId: roktTagId,
             sdkVersion: libraryVersion,
             sessionManager: TxnSessionManager(roktTagId: roktTagId),
-            httpClient: httpClient
+            httpClient: httpClient,
+            deviceHeaders: NetworkingHelper.txnDeviceHeaders()
```
alexs-mparticle, line 976: `txnDeviceHeaders()` sets `rokt-package-name` only when `Bundle.main.bundleIdentifier` resolves (absent otherwise), while v1 `getCommonHeaders` force-unwraps it. The load-bearing header (gateway page detection keys off it) now depends on this wiring; `deviceHeaders` defaults to `[:]`. Correct in this one production path, so latent — but the always-present guarantee is gone.

Sources/Rokt_Widget/Networking/OffersClient.swift
```diff
@@ -39,22 +42,33 @@ internal struct OffersClient {
...
             "Content-Type": "application/json"
         ]
+        // Authorization is optional: with no live token the server mints a fresh session.
+        if let authToken, !authToken.isEmpty {
```
alexs-mparticle: `!authToken.isEmpty` is dead — `authorizationHeader` returns nil or `"Bearer …"`, never an empty non-nil string. Simplify to `if let authToken`.

alexs-mparticle (commented):
Review — mostly around the shared real-time-event store, plus cleanup. Wire-shape changes look solid and well-tested.

Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
crisryantan: Fixed — capped the untriggered store at `maximumRealTimeEventsToStore` (evict oldest) in both the file and memory stores, so distinct echoed events no longer accumulate. Added a cap test. https://github.com/ROKT/rokt-sdk-ios/commit/1c3b96b

Sources/Rokt_Widget/DataAccess/RealtimeEventStore.swift
crisryantan: Fixed — `addUntriggeredEvents` now runs inside `eventProcessingQueue` (the same serial queue `markAsTriggered` uses), so the read-modify-write can't lose updates under concurrent captures. https://github.com/ROKT/rokt-sdk-ios/commit/1c3b96b

Sources/Rokt_Widget/Networking/OffersService.swift
crisryantan: Left as-is for this PR, deliberately. This mirrors the v1 path exactly (same process-wide store, cleared only on `sessionInvalidated()`), so it's not a regression introduced here, and the real-time events carry their own parent/instance guids. Agree it's worth confirming gateway idempotency on cross-session re-sends — and session-scoping the store (or clearing triggered events when the session id changes) is a sensible follow-up that I didn't want to fold a store-lifecycle change into this PR for. Flagging it for the v2 events follow-up.

Sources/Rokt_Widget/Networking/SelectEventMapper.swift
crisryantan: Done — extracted `UntriggeredRealTimeEvent.flattenedValid` over a small `RealTimeEventSignal` protocol; the v1 `UntriggeredEventsContainer` and v2 `SelectEventMapper` now flatten/validate through it, so the validity rule lives in one place. They keep separate Decodables only for the camelCase/snake_case split. https://github.com/ROKT/rokt-sdk-ios/commit/1c3b96b

Sources/Rokt_Widget/Networking/OffersClient.swift
crisryantan: Done — added `TxnRequestHeaders.common(accountId:authToken:)` as the single owner for account-id / content-type / rokt-txn-shadow / optional Authorization; OffersClient, TxnEventsClient and TxnInitClient all route through it, so toggling or renaming shadow/auth is a one-line change in one place. https://github.com/ROKT/rokt-sdk-ios/commit/1c3b96b

Sources/Rokt_Widget/Networking/OffersService.swift
crisryantan: Fixed — `getExperienceData` computes `getPrivacyControlPayload` once and passes the payload into both `buildPrivacyControl` and `buildPrivacy`. https://github.com/ROKT/rokt-sdk-ios/commit/1c3b96b

Sources/Rokt_Widget/RoktInternalImplementation.swift
crisryantan: Kept as-is, deliberately: omit-when-absent is safer than v1's force-unwrap (which crashes if `bundleIdentifier` is ever nil), and the bundle id is effectively always present for a shipping app. The one production path (`defaultOffersService` -> `txnDeviceHeaders()`) always supplies it, and the offers pact + the gateway's missing-package-name 409 are the guardrails. Happy to add an assert/log if you'd prefer a louder signal on the latent case.

Sources/Rokt_Widget/Networking/OffersClient.swift
crisryantan: Fixed — simplified to `if let authToken` (the `!isEmpty` was dead; `authorizationHeader` returns nil or a Bearer string), now living once inside `TxnRequestHeaders.common`. https://github.com/ROKT/rokt-sdk-ios/commit/1c3b96b

alexs-mparticle (approved):

PR #216, opened 2026-06-26
feat(offers): forward real-time events on the v2 offers request

crisryantan:
## Summary

Stacked on #215. Closes the last v2 offers request-shape gap versus Android: real-time events triggered during the previous placement are now forwarded on `POST /v2/sessions/offers`, and the response's `event_data` is captured for the next call.

## Changes

**Request `events[]`.** Forwarded events reuse `TxnEvent` — the same element as `/v2/sessions/events` (`event_type`, epoch-ms `timestamp`, `data.payload`; `instance_id` omitted) — matching Android. The new `SelectEventMapper` maps the real-time event store's triggered events into that shape, parsing the stored event time to epoch-ms via `EventDateFormatter` (same as `TxnEventMapper`).

**Session-gated.** Events are forwarded only when there's a live session token to attribute them to (mirrors Android); the key is omitted when there are none.

**Response capture.** On a successful response, `event_data` is flattened into untriggered events and added to the store for the next placement. It only adds — no global clear — mirroring the v1 experiences capture, so the process-wide `RealTimeEventManager` (shared with the events and v1 paths) isn't wiped.

## Notes

The new `events[]` rides on the live request but is not asserted in the consumer pact yet, so the published contract is unchanged.

## Testing

`xcodebuild test` on the iPhone 17 Pro simulator — new `TestSelectEventMapper` (mapper round-trip + fallbacks), `TestSelectSerialization` (events element shape + omit-when-absent), `TestOffersService` (forward only with a live session; response-capture); `OffersClientPactSpec` unchanged and green.

### Verified end-to-end (live traffic)

Captured two consecutive offers requests from one live session in the iOS example app (credentials/payloads redacted):

1. **First selection** — no `events` key (nothing triggered yet).
2. **Second selection** (~28s later, after the placement rendered and fired signals) — `events[]` with 12 entries, each in the `/v2/sessions/events` element shape:

   ```json
   { "event_type": "PlacementImpressionV2", "timestamp": 1782487352257, "data": { "payload": "<base64 realtime-event protobuf>" } }
   ```

   - epoch-ms `timestamp`, no `instance_id` (matches Android)
   - every timestamp falls between the two requests — i.e. the events fired while the first placement was on screen
   - decoded payloads are `realtimeevents.v1.*PayloadData` protobufs echoed verbatim from the prior response's `event_data`, confirming the response → store → forward capture-back loop
   - 12 events, under the 50 cap

A fresh app relaunch started clean (no stale events on the first request), and the device headers from #215 were present throughout.

### DIFF
--- a/Sources/Rokt_Widget/Models/SelectResponse.swift
+++ b/Sources/Rokt_Widget/Models/SelectResponse.swift
@@ -16,8 +16,8 @@ internal struct SelectRequest: Encodable, Equatable {
     let privacyControl: SelectPrivacyControl?
     let privacy: SelectPrivacy?
     // Real-time events triggered during the previous placement, forwarded for the next
-    // selection. Reuses TxnEvent (the /v2/sessions/events element); omitted when none.
-    let events: [TxnEvent]?
+    // selection; omitted when none.
+    let events: [SelectEvent]?
 
     enum CodingKeys: String, CodingKey {
         case page
@@ -34,7 +34,7 @@ internal struct SelectRequest: Encodable, Equatable {
         attributes: [String: String] = [:],
         privacyControl: SelectPrivacyControl? = nil,
         privacy: SelectPrivacy? = nil,
-        events: [TxnEvent]? = nil
+        events: [SelectEvent]? = nil
     ) {
         self.page = page
         self.channel = channel
@@ -113,6 +113,33 @@ internal struct SelectPrivacy: Encodable, Equatable {
 
     init(gpcEnabled: Bool? = nil) {
         self.gpcEnabled = gpcEnabled
+    }
+}
+
+/// A real-time event forwarded on the offers request: the provider's typed event
+/// plus its echoed `payload`, stamped with an epoch-ms `timestamp`. Encodes to the
+/// `/v2/sessions/events` element shape (`event_type`, `timestamp`, `data.payload`).
+internal struct SelectEvent: Encodable, Equatable {
+    let eventType: String
+    let timestamp: Int64
+    let payload: String
+
+    enum CodingKeys: String, CodingKey {
+        case eventType = "event_type"
+        case timestamp
+        case data
+    }
+
+    private enum DataKeys: String, CodingKey {
+        case payload
+    }
+
+    func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(eventType, forKey: .eventType)
+        try container.encode(timestamp, forKey: .timestamp)
+        var data = container.nestedContainer(keyedBy: DataKeys.self, forKey: .data)
+        try data.encode(payload, forKey: .payload)
     }
 }
 
--- a/Sources/Rokt_Widget/Models/SelectResponse.swift
+++ b/Sources/Rokt_Widget/Models/SelectResponse.swift
@@ -15,6 +15,9 @@ internal struct SelectRequest: Encodable, Equatable {
     let attributes: [String: String]
     let privacyControl: SelectPrivacyControl?
     let privacy: SelectPrivacy?
+    // Real-time events triggered during the previous placement, forwarded for the next
+    // selection. Reuses TxnEvent (the /v2/sessions/events element); omitted when none.
+    let events: [TxnEvent]?
 
     enum CodingKeys: String, CodingKey {
         case page
@@ -22,6 +25,7 @@ internal struct SelectRequest: Encodable, Equatable {
         case attributes
         case privacyControl = "privacy_control"
         case privacy
+        case events
     }
 
     init(
@@ -29,13 +33,15 @@ internal struct SelectRequest: Encodable, Equatable {
         channel: SelectChannel,
         attributes: [String: String] = [:],
         privacyControl: SelectPrivacyControl? = nil,
-        privacy: SelectPrivacy? = nil
+        privacy: SelectPrivacy? = nil,
+        events: [TxnEvent]? = nil
     ) {
         self.page = page
         self.channel = channel
         self.attributes = attributes
         self.privacyControl = privacyControl
         self.privacy = privacy
+        self.events = events
     }
 }
 
--- a/Sources/Rokt_Widget/Networking/OffersService.swift
+++ b/Sources/Rokt_Widget/Networking/OffersService.swift
@@ -127,8 +127,11 @@ internal struct OffersService {
             deviceHeaders: deviceHeaders,
             httpClient: httpClient
         )
-        // Forward events triggered during the previous placement; read once, before retries.
-        // Only with a live session to attribute them to, matching Android.
+        // Forward events triggered during the previous placement; read once, before retries,
+        // and only with a live session to attribute them to (matching Android). As on the v1
+        // path, triggered events are not cleared after forwarding — they ride subsequent
+        // requests until session invalidation; re-send is expected (the "only adds, no clear"
+        // note elsewhere is about the untriggered response-capture, not this read).
         let forwardedEvents = authToken != nil ? SelectEventMapper.requestEvents(from: triggeredEvents()) : []
         let input = OffersInput(
             requestId: makeRequestId(),
--- a/Sources/Rokt_Widget/Networking/OffersService.swift
+++ b/Sources/Rokt_Widget/Networking/OffersService.swift
@@ -24,6 +24,11 @@ internal struct OffersService {
     let makeRequestId: () -> String
     let makePageInstanceGuid: () -> String
     let completionQueue: DispatchQueue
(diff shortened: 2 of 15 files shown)


Sources/Rokt_Widget/Models/SelectResponse.swift
```diff
@@ -15,27 +15,33 @@ internal struct SelectRequest: Encodable, Equatable {
     let attributes: [String: String]
     let privacyControl: SelectPrivacyControl?
     let privacy: SelectPrivacy?
+    // Real-time events triggered during the previous placement, forwarded for the next
+    // selection. Reuses TxnEvent (the /v2/sessions/events element); omitted when none.
+    let events: [TxnEvent]?
```
alexs-mparticle: I thought we were removing `Txn` prefix?

crisryantan: Good call — swapped it for a prefix-free `SelectEvent` (same wire shape, just the fields the forward needs), so the offers model no longer pulls in a `Txn`-named type. https://github.com/ROKT/rokt-sdk-ios/commit/135247d

Sources/Rokt_Widget/Networking/OffersService.swift
```diff
@@ -143,6 +161,10 @@ internal struct OffersService {
                 let decoded = try JSONDecoder().decode(SelectResponse.self, from: data)
                 // Roll the refreshed token forward for the next offers/events call.
                 await sessionManager.update(sessionToken: decoded.sessionToken)
+                // Capture the echoed events so the next placement can forward them back.
+                if let eventData = decoded.eventData {
+                    captureEvents(SelectEventMapper.untriggeredEvents(from: eventData))
```
alexs-mparticle, line 169: **Capturing into an unbounded store.** `captureEvents` → `RealTimeEventManager.addUntriggeredEvents` does a full read-append-rewrite on every successful offers response, and only the *triggered* file is trimmed (`trimTriggeredEvents`, 50-cap) — the untriggered file has no cap. If the backend echoes overlapping `event_data` across consecutive responses (same placement/offer is common), `untriggered_events.json` accumulates duplicate rows for the life of the session, with O(n) re-encode per response. Worse, when one trigger later matches, `updateTriggeredEvents` emits one `TriggeredRealTimeEvent` per duplicate row, multiplying what gets forwarded. Consider trimming the untriggered file the way the triggered file is trimmed, and/or de-duping on append.

Sources/Rokt_Widget/Networking/OffersService.swift
```diff
@@ -104,21 +117,26 @@ internal struct OffersService {
...
-            authToken: await sessionManager.authorizationHeader,
+            authToken: authToken,
             sdkVersion: sdkVersion,
             pageInstanceGuid: makePageInstanceGuid(),
             deviceHeaders: deviceHeaders,
             httpClient: httpClient
         )
+        // Forward events triggered during the previous placement; read once, before retries.
+        // Only with a live session to attribute them to, matching Android.
+        let forwardedEvents = authToken != nil ? SelectEventMapper.requestEvents(from: triggeredEvents()) : []
```
alexs-mparticle, line 135: **Triggered events are read but never cleared/marked-sent**, so the same events ride every subsequent offers request until `sessionInvalidated()`. Note this mirrors the pre-existing v1 path (`addRealtimeEventsIfPresent`), so it's established design / Android parity rather than a regression here — flagging only because the PR description's "only adds, no global clear" rationale is about the *untriggered* response-capture path, not this *triggered*-forward read. A one-line comment clarifying that, or a note that the backend dedupes, would prevent a future reader from assuming this path was vetted for re-send.

Sources/Rokt_Widget/Networking/SelectEventMapper.swift
```diff
@@ -0,0 +1,50 @@
...
+                eventType: event.eventType,
+                timestamp: epochMilliseconds(from: event.eventTime),
+                payload: event.payload
+            )
+        }
+    }
+
+    /// Response `event_data` → untriggered events for the store, consulted when the
+    /// next placement fires (mirrors ``UntriggeredEventsContainer``'s flattening).
+    static func untriggeredEvents(from eventData: [String: SelectEventDataEntry]) -> [UntriggeredRealTimeEvent] {
```
alexs-mparticle, line 28: **Duplicated flattening + dropped validity filter.** This loop re-implements `UntriggeredEventsContainer.init(from:)` (UntriggeredRealTimeEvent.swift:20-43) — same parentGuid→signalKey→`UntriggeredRealTimeEvent` construction over the same `event_data` envelope, and `SelectEventDataEntry`/`SelectRealTimeEvent` duplicate the private `RawEventData`/`RawEvent`. A wire-schema change now has to be applied in two places or the two ingestion paths silently diverge. Consolidating (have `UntriggeredEventsContainer` decode `[String: SelectEventDataEntry]` and delegate here) would delete ~40 lines.

Separately, unlike `UntriggeredEventsContainer` this does **not** apply the `isValid()` filter at construction. Harmless today because the only caller re-filters via `RealTimeEventManager.addUntriggeredEvents`, but it's an undocumented divergence from the sibling it mirrors — worth either matching the filter or a one-line comment that validation is delegated to the store.

Sources/Rokt_Widget/Networking/SelectEventMapper.swift
```diff
@@ -0,0 +1,50 @@
...
+                    )
+                )
+            }
+        }
+        return events
+    }
+
+    // Mirrors TxnEventMapper: parse the stored ISO event time to epoch-ms, falling back to now.
+    private static func epochMilliseconds(from eventTime: String) -> Int64 {
+        if let date = EventDateFormatter.dateFormatter.date(from: eventTime) {
```
alexs-mparticle: `epochMilliseconds(from:)` is byte-identical to `TxnEventMapper.epochMilliseconds` (TxnEventMapper.swift:125-130) — same parse-or-fallback-to-now over the same `EventDateFormatter.dateFormatter`. Extract one shared helper (a static on `EventDateFormatter`, which already owns the formatter) so the fallback policy can't drift between the offers and txn event paths.

alexs-mparticle (commented):
Automated review (8 finder angles + per-finding verification). The wire shape and tests look solid — these comments are about shared-store semantics and some duplication. Two widely-suggested issues were refuted and dropped: the `authToken != nil` gate already excludes expired sessions (`authorizationHeader` returns nil when `hasExpired`), and the epoch-ms parse-or-fallback is intentional, tested parity with `TxnEventMapper`. The two I'd act on before merge are the unbounded untriggered file and the event_data duplication.

Sources/Rokt_Widget/Networking/OffersService.swift
crisryantan: Fixed — `addUntriggeredEvents` now de-dupes on append in both the file and memory stores, so identical `event_data` echoed across responses no longer accumulates or multiplies on trigger-match. Added a store test that proves one trigger yields one event after repeated identical adds. https://github.com/ROKT/rokt-sdk-ios/commit/5d98527

Sources/Rokt_Widget/Networking/OffersService.swift
crisryantan: Added a comment at the forward read clarifying that triggered events aren't cleared after forwarding (mirrors v1; they ride subsequent requests until session invalidation, re-send expected) and that the "only adds, no clear" note is about the untriggered response-capture, not this read. https://github.com/ROKT/rokt-sdk-ios/commit/5d98527

Sources/Rokt_Widget/Networking/SelectEventMapper.swift
crisryantan: Added the `isValid()` filter so it matches `UntriggeredEventsContainer` rather than relying on the store to drop invalids. Kept the flatten separate, though: v1 decodes camelCase `eventType` (`RawEvent`) and v2 decodes snake_case `event_type` (`SelectRealTimeEvent`), so the decode types can't merge without breaking one — and coupling the canonical v2 path back to the legacy decode felt like the wrong direction. Left a comment noting the parallel + the filter. https://github.com/ROKT/rokt-sdk-ios/commit/5d98527

Sources/Rokt_Widget/Networking/SelectEventMapper.swift
crisryantan: Extracted `EventDateFormatter.epochMilliseconds(from:)` (it already owns the formatter); both `SelectEventMapper` and `TxnEventMapper` use it now, so the parse-or-fallback policy lives in one place. https://github.com/ROKT/rokt-sdk-ios/commit/5d98527

alexs-mparticle (approved):

PR #226, opened 2026-06-30
feat(txn): migrate iOS init to config-only GET /v2/init

jaissica12:
## Background

Move the v2 init call from `POST /v2/sessions/init` to a body-less
`GET /v2/init` (all inputs as headers) that returns **config only** —
`feature_flags` + `fonts`, with no `session_id` / `session_token`.

`POST /v2/sessions/init` previously returned config **and** minted a per-client
`session_id` + JWT `session_token` in one response. Carrying a per-client secret
made the response impossible to edge-cache safely (a partner/platform-keyed cache
would share one user's token across the whole audience). Per the init/font
alignment decision, init becomes config-only and cacheable, and the session is
now sourced from the first `offers/select` response (the v1 shape) — v2 offers
already mints + returns a session token when none is sent.

The endpoint is named `/v2/init` to match the SDK's public
`Rokt.init` surface, per the alignment decision. 

**Dependencies:** requires tx-api to serve `GET /v2/init`
(`transactions#feat/txn-init-rename-v2-init`) and Akamai to route it
(`rokt-akamai#feat/v2-init-get`, on PR #350). Keep in lockstep with the Android
change (`sdk-android-source#feat/v2-init-get`).

Fixes [(issue)]

## What Has Changed

- **`TxnInitClient`** — migrated from POST to body-less `GET /v2/init`; inputs
  passed as explicit headers, no request body, no `Authorization`. Removed the
  now-unused `authToken`, `TxnInitRequest`, and `TxnInitClientError`.
- **`TxnInitResponse`** — config-only: dropped `session_id` / `session_token`
  (keeps `feature_flags` + `fonts`).
- **`TxnInitService`** — updated to the config-only GET call shape (no token
  input).
- **`RoktInternalImplementation` / `TestTxnInitWiring`** — removed the init
  session/token wiring that no longer applies.
- **Tests / mocks** — `TxnInitClientPactSpec`, `TestTxnInitService`,
  `TestMockTxnInitHTTPClient`, `TestTxnInitResponse`,
  `TestTxnInitResponseAdapter`, and `MockResponse` updated to the
  GET `/v2/init` config-only, header-based shape.

## How Has This Been Tested

- Updated and ran the unit tests for the init client/service/response
  (`TestTxnInitService`, `TestTxnInitResponse`, `TestMockTxnInitHTTPClient`).
- Updated the consumer contract test (`TxnInitClientPactSpec`) to assert the
  body-less `GET /v2/init` request and config-only response; verified the Pact
  regenerates against the tx-api provider state.

## Notes

- Fonts ride the config-only response and are populated server-side from the
  op3 `AccountIntegration` projection — no iOS change needed for fonts.
- Session sourcing already exists in the offers/select path, so init minting
  nothing does not change the public init API behaviour for integrators.

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/Networking/TxnInitClient.swift
+++ b/Sources/Rokt_Widget/Networking/TxnInitClient.swift
@@ -27,8 +27,7 @@ internal struct TxnInitClient {
     ) async throws -> (Data?, HTTPURLResponse?) {
         let url = baseURL
             .appendingPathComponent("v2")
-            .appendingPathComponent("sessions")
-            .appendingPathComponent("init")
+            .appendingPathComponent("config")
 
         let requestBody = TxnInitRequest(
             operatingSystem: operating_system,
--- a/Sources/Rokt_Widget/Networking/TxnInitClient.swift
+++ b/Sources/Rokt_Widget/Networking/TxnInitClient.swift
@@ -18,12 +18,8 @@ internal struct TxnInitClient {
         self.httpClient = httpClient
     }
 
-    /// Calls the config-only `GET /v2/init`.
-    ///
-    /// Every input travels as a request header — there is no body and no
-    /// `Authorization` — so the request is a plain, cacheable GET. The response
-    /// carries no session; the SDK sources its session from the offers/select
-    /// response instead.
+    /// Config-only `GET /v2/init`: inputs as headers, no body/auth, no session
+    /// in the response (session comes from offers/select).
     func initSession(
         operating_system: String,
         layout_schema_version: String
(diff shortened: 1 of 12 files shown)


Sources/Rokt_Widget/Networking/TxnInitClient.swift
```diff
@@ -3,51 +3,48 @@ import Foundation
...
         self.baseURL = baseURL
         self.accountId = accountId
-        self.authToken = authToken
         self.sdkVersion = sdkVersion
         self.httpClient = httpClient
     }
 
+    /// Calls the config-only `GET /v2/init`.
+    ///
+    /// Every input travels as a request header — there is no body and no
```
crisryantan: yeah lets clean this up

jaissica12: done

crisryantan (approved):

denischilik (approved):

darrenli-rokt (approved):

PR #229, opened 2026-07-02
fix: Retryable Stripe Support

BrandonStalnaker:
## Background

- Built-in device pay (including Stripe-routed card flows) reports three distinct outcomes from the payment orchestrator: success, user cancellation, and failure. Previously, any non-success result was treated as a failure: the SDK always called devicePayFinalized, which closed the UX flow, and cancellation could not be retried.
- When a user dismisses the Stripe payment sheet without completing checkout, that should be a recoverable cancellation—not a terminal failure. The UX layer exposes devicePayRetry for this case, but the SDK was not wiring it up from the payment completion handler.
- Failure callbacks also omitted the underlying error message (error: nil), making it harder for integrators to diagnose payment issues.

## What Has Changed

- Added a devicePayRetry helper that forwards cancellation to the UX helper so the user can attempt payment again without finalizing the flow.
- Replaced the binary success/failure check in the device pay completion handler with explicit handling per outcome:
- Succeeded — emits CartItemInstantPurchase and calls devicePayFinalized(success: true).
- Canceled — calls devicePayRetry instead of finalizing (no failure event).
- Failed — emits CartItemInstantPurchaseFailure with result.errorMessage and calls devicePayFinalized(success: false).
- Unknown — same as failed for forward compatibility.

## How Has This Been Tested

Describe the tests that you ran to verify your changes.

## Notes

The branch pin in Package.swift is required until the UX helper PR merges and releases. Coordinate both PRs, then bump the dependency to the released version.

## Checklist

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,7 +21,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/ROKT/rokt-contracts-apple.git", .upToNextMajor(from: "2.0.2")),
-        .package(url: "https://github.com/ROKT/rokt-ux-helper-ios.git", branch: "fix/Retryable-Stripe-Support"),
+        .package(url: "https://github.com/ROKT/rokt-ux-helper-ios.git", from: "0.13.0"),
         .package(url: "https://github.com/WeTransfer/Mocker.git", .upToNextMajor(from: "2.0.0")),
         .package(url: "https://github.com/surpher/PactSwift.git", .upToNextMajor(from: "1.2.0"))
     ],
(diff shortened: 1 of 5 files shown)


Package.swift
```diff
@@ -21,7 +21,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/ROKT/rokt-contracts-apple.git", .upToNextMajor(from: "2.0.2")),
-        .package(url: "https://github.com/ROKT/rokt-ux-helper-ios.git", from: "0.12.0"),
+        .package(url: "https://github.com/ROKT/rokt-ux-helper-ios.git", branch: "fix/Retryable-Stripe-Support"),
```
BrandonStalnaker: Do not merge with this. Its just to allow the tests to pass while the new version of the ux-helper is not released.

denischilik (approved):

PR #230, opened 2026-07-06
fix(init): make font loading non-blocking on the transactions init path

jaissica12:
## Background

- The legacy init flow coupled custom-font handling to initialization: `InitComplete` fired only *after* fonts downloaded, the first placement render waited on font load state, and a single font failure marked the whole SDK "not initialized" (dropping subsequent `execute` calls).

## What Has Changed

Behavior (transactions init path):
- Init now completes on **config** — `InitComplete` fires immediately, no longer gated on font download. Any `execute` that arrived before init finished is replayed right away.
- Fonts load **off the critical path**; renders use system fonts and custom fonts swap in when they arrive.
- Font failures are **best-effort**: reported as diagnostics only. The SDK stays initialized and falls back to system fonts (removed the "un-initialize on font error" behavior).

Isolation (so the legacy path stays cleanly removable):

- **iOS**: new `handleTxnInitSuccess`; the legacy `handleInitSuccess`, font observers, and `showNow` font gate are untouched and wired **only** on the legacy branch.

Files — iOS:
- `RoktInternalImplementation.swift` (new `handleTxnInitSuccess`, observers on legacy branch only), `Util/FontManager.swift`, `Networking/RoktNetworkAPI.swift` (fail-soft: no longer un-initialize on font error)

## Screenshots/Video

- N/A (no visual change; behavioral — init/render no longer wait on fonts)

## Checklist

- [ ] I have tested this locally.

## Additional Notes

- The legacy (v1) init path is byte-for-byte untouched and self-contained for clean removal.
- The shared `FontManager`/`RoktNetworkAPI` keep one low-level fix — a font failure no longer un-initializes the SDK — which is required by the transactions path, harmless to the (unbound) legacy path, and survives v1 removal.
- Out of scope (pre-existing, follow-up): font cache-key stability/eviction, download retry/backoff, and a proactive "fonts ready" re-render.

## Reference Issue (For employees only. Ignore if you are an outside contributor)

- Closes [ticket](https://go/j/[ticket-number])


Sources/Rokt_Widget/Networking/RoktNetworkAPI.swift
```diff
@@ -175,7 +175,7 @@ internal class RoktNetWorkAPI {
 
                     return
                 } else {
-                    Rokt.shared.roktImplementation.isInitialized = false
+                    // Best-effort: mark for diagnostics, don't un-initialise.
```
crisryantan, line 179: do we need the comments?

jaissica12: cleaning up

nickolas-dimitrakas (approved):

PR #243, opened 2026-07-15
fix(txn): drop session and re-mint on 401 (token validation)

nickolas-dimitrakas:
## Why

tx-api is enabling **token validation** (Monitoring → Enforcement for mSDK/wSDK/S2S). On a rejected token the gateway returns **401**. The SDK must not keep reusing a rejected token (that loops on 401s), and it must not silently lose events.

## What the SDK does on a 401

**Offers — `POST /v2/sessions/offers`:**
- A 401 from offers is **not expected** — the gateway mints a fresh session even with no token. Defensive handling: on 401, **drop the stored session** and surface the failure for this placement. The next offers call sends no `Authorization` and re-mints. (No inline retry — offers is itself the mint endpoint.)

**Events — `POST /v2/sessions/events`:**
- On 401, **drop the batch, clear the session, and emit a `[TXN_EVENTS_401]` diagnostic.** A 401 indicates a forged/corrupted token (exceptional); it is not recoverable, and re-minting would attach the events to a new, *unlinked* session (orphaning them). The recoverable case (an expired token) returns `200` with a fresh token bound to the **same** session id and is handled by the normal success path — it never reaches this 401 branch.

**Init** (`GET /v2/init`) is config-only with no auth, so it's unaffected.

## API contract this relies on
Under JWT enforcement: an **expired** (or unknown-kid) token returns `200` — offers mints a new session, events returns a fresh token bound to the same session id — so those recover on the normal success path. A `401` is returned only for a forged/corrupted token, which the SDK drops.

## Diagnostics
Full diagnostics wiring is per the SDK team; the events drop-after-re-mint emits a `[TXN_EVENTS_401]` diagnostic, and both paths log (verbose on the normal self-heal, error on an unexpected persistent failure).

## Tests
- `TestOffersService.test_getExperienceData_onUnauthorized_dropsSessionAndFails` — offers 401 drops session + surfaces failure.
- `TestTxnEventService.test_send_unauthorized_dropsSessionAndDoesNotRetry` — events 401 drops the batch, clears the session, no retry.

All 28 tests across the two suites pass (`TEST SUCCEEDED`).

## Files
- `Sources/Rokt_Widget/Networking/OffersService.swift`
- `Sources/Rokt_Widget/Networking/TxnEventService.swift`
- `Tests/Rokt_WidgetTests/Shared/Networking/{TestOffersService,TestTxnEventService}.swift`


Sources/Rokt_Widget/Networking/OffersService.swift
```diff
@@ -124,38 +124,64 @@ internal struct OffersService {
...
         let forwardedEvents = authToken != nil ? SelectEventMapper.requestEvents(from: triggeredEvents()) : []
-        let input = OffersInput(
-            requestId: makeRequestId(),
-            pageIdentifier: pageIdentifier,
-            attributes: attributes,
-            privacyControl: privacyControl,
-            privacy: privacy,
-            events: forwardedEvents.isEmpty ? nil : forwardedEvents
-        )
+        func makeInput(includeForwardedEvents: Bool) -> OffersInput {
```
denischilik: Could you please move function on class level?

Sources/Rokt_Widget/Networking/TxnEventService.swift
```diff
@@ -79,6 +79,14 @@ internal struct TxnEventService {
                     continue
                 }
 
+                // Token rejected: drop the stale session so the next offers call re-mints a
+                // fresh one. Events are best-effort, so we don't retry the batch.
+                if statusCode == HTTPStatusCode.unauthorized.rawValue {
+                    RoktLogger.shared.verbose("Events returned 401; dropping session")
+                    await sessionManager.clear()
+                    throw TxnEventError.unexpectedStatusCode(statusCode)
```
jamesnrokt: Won't we need retry logic here? Dropping events and data is a big warning flag to me and want to make sure we're aligned with the backend expectations

Sources/Rokt_Widget/Networking/OffersService.swift
```diff
@@ -124,38 +124,64 @@ internal struct OffersService {
...
         while true {
             do {
                 let (data, response) = try await client.fetchOffers(input: input)
                 let statusCode = response?.statusCode ?? 0
 
+                // Token rejected (server-side validation/revoked/expired): drop the stale
+                // session and re-mint once with no Authorization — the gateway issues a fresh
+                // session when no token is present, so the placement self-heals instead of
+                // looping on an invalid token.
+                if statusCode == HTTPStatusCode.unauthorized.rawValue, !didReMintAfterUnauthorized {
```
denischilik: Could you please add "dedossing" scenario

denischilik (commented):
Some comments

Sources/Rokt_Widget/Networking/TxnEventService.swift
BrandonStalnaker: I don't think so. I don't think retry is possible without offers orchestration so session clear is the right hygiene step I think. Nick?

Sources/Rokt_Widget/Networking/OffersService.swift
nickolas-dimitrakas: Done in a1466c5 — hoisted both helpers out of `fetchExperienceString` into private class-level methods `makeOffersClient(baseURL:pageInstanceGuid:authToken:)` and `makeOffersInput(...)`. Per-placement values (baseURL, pageInstanceGuid, requestId, forwardedEvents) are passed as params. No behavior change; tests still green.

Sources/Rokt_Widget/Networking/OffersService.swift
nickolas-dimitrakas: Yep, added in e377c1d

Sources/Rokt_Widget/Networking/TxnEventService.swift
nickolas-dimitrakas: Good flag — updated in 225dc41 so we no longer drop the events. On a 401 we clear the (dead) session and **buffer** the undelivered events in a `TxnPendingEventsStore`; the next `send` drains the buffer and re-sends them first, by which point an offers call has typically minted a fresh session/token, so they transfer to it. Buffer is capacity-bounded so it can't grow unbounded if a session is never minted.

Retrying inline isn't possible without offers orchestration (the events endpoint can't mint a session — only `/v2/sessions/offers` can), so this "hold and re-send under the next session" is the equivalent of a retry for the events path — matches what Brandon noted.

One thing I want to confirm with SYD before we consider this final: re-sending events that originated under the now-invalidated session under a **newly minted** session assumes the backend attributes them as intended (no double-count / mis-attribution). Raising that with them now — will update here once confirmed.

nickolas-dimitrakas: @jamesnrokt @BrandonStalnaker Closing the loop: confirmed — the events endpoint can't mint a session and we don't want old-session events re-attributed to a new one, so we drop the events + clear the session, with a `[TXN_EVENTS_401]` diagnostic for visibility. Brandon's read was right — session clear is the correct hygiene step here; done in 2109107.

denischilik (approved):

BrandonStalnaker (approved):

PR #251, opened 2026-07-17
feat(txn): persist unsent event batches and replay on next init

nickolas-dimitrakas:
## What

Adds a persisted offline replay queue for transaction events, aligning with the web SDK's `RoktTransactionsPendingEvents` store.

- **`TxnPendingEventStore`** (new): disk-backed (Caches dir), **cap 10 batches, 30-min TTL**, newest dropped when full (matches web `_persistRequest`). Expired batches don't consume capacity.
- **`TxnEventService`**: on a *terminal recoverable* failure (exhausted 5xx / transient transport), the batch is persisted instead of dropped. Permanent failures (**400/401**) are not replayed.
- **Replay on init**: `replayPendingTxnEvents()` drains non-expired batches and re-dispatches them; invoked from `handleInitSuccess`. Drained batches that fail again are re-persisted.
- `TxnEvent` / `TxnEventDataValue` are now `Codable` so batches round-trip to disk.

## Why

Previously `dispatchTxnEvents` did `Task { try? await service.send(...) }` — any post-retry failure was silently dropped. This was the largest caching divergence from web in the parity analysis.

## Design note

Web persists each request *while in flight* and removes it on completion (kill-resilience via `sessionStorage`). iOS here **enqueues on terminal failure** — this matches the sdk-android PR4 parity plan and is lower-I/O; observable replay behavior (offline/rate-limited batches survive and replay next init, 10/30-min) is equivalent. Dedup of persisted batches is intentionally not mirrored (web dedups by `JSON.stringify(body)`); called out for review.

## Tests

- `TestTxnPendingEventStore` (7): persist/drain, clear-on-drain, empty no-op, TTL drop, TTL boundary, cap-of-10 drops-newest, expired-don't-consume-capacity.
- `TestTxnEventService` (4 new): persists on exhausted 503; does **not** persist on 400 / 401 / success.

### DIFF
--- a/Sources/Rokt_Widget/DataAccess/TxnPendingEventStore.swift
+++ b/Sources/Rokt_Widget/DataAccess/TxnPendingEventStore.swift
@@ -1,0 +1,76 @@
+import Foundation
+
+/// A single unsent events batch awaiting replay.
+internal struct TxnPendingEventBatch: Codable, Equatable {
+    let events: [TxnEvent]
+    let persistedAtMs: Int64
+}
+
+/// Persists unsent event batches so they survive a process restart and can be replayed on the
+/// next init. Mirrors the web SDK's `RoktTransactionsPendingEvents` store: at most 10 batches,
+/// each with a 30-minute TTL.
+internal protocol TxnPendingEventStoring {
+    /// Persists a batch that failed to send. No-op once the store is full (newest dropped).
+    func persist(events: [TxnEvent])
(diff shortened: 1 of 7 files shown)


jamesnrokt (approved):
Approved but recommending testing in conjunction with #250 as we will be re-writing the event store from the background. You've got defensive try's in place so it should be enough but warrants further investigation

PR #260, opened 2026-07-20
perf(txn): collapse Select*/Render* offers double-model into one encoder

nickolas-dimitrakas:
## Summary
Third in the stacked SDK-size series and the **biggest lever** for the +292 KB v2-transactions regression. **Stacked on #259** (which is stacked on #258).

The offers path decoded `SelectResponse`, mapped it into a parallel `Render*` model tree (16 `Encodable` structs), and re-encoded that to the camelCase experience JSON string `RoktUX.parseExperience` consumes. That tree was **~88 KiB** of the binary — **~48 KiB of it pure type metadata/witnesses** for the 16 types.

This replaces the tree with a single `RenderExperience: Encodable` that wraps the decoded `SelectResponse` and emits the identical JSON in one pass via `KeyedEncodingContainer` + private helpers (one dynamic `CodingKey`, no per-shape enums). Same `JSONEncoder(.sortedKeys)`, so number formatting and key ordering are unchanged; the required-vs-`encodeIfPresent` field-presence rules are preserved exactly.

**No `rokt-ux-helper-ios` change needed** — RoktUX only ingests a JSON string (`parseExperience`), so the contract is unchanged. (The `Select*` decode tree still decodes the wire, so a model-based RoktUX API wouldn't add binary savings.)

## Impact
- Release SDK executable **−65,544 bytes (~64 KiB)**.
- Combined with #258 + #259: **−83,640 bytes** off the txn merge (~28% of the +292 KB regression recovered).

## Automated validation
- **New `TestOffersRenderGolden`** compares the encoder output to goldens captured from the previous model-tree implementation (committed as `*.expected.json`), via deep parsed-JSON equality — catches any changed key, value, nesting, or spurious `null`.
- `TestSelectExperienceAdapter` (structural contract), `TestOffersRenderCompat` (`RoktUX.parseExperience` end-to-end, incl. shoppable catalog), `TestSelectSerialization`, `TestSelectEventMapper` — **22/22 pass**.
- CI: SPM Unit Tests, Periphery Scan, Trunk Check green; SDK Size Report shows **4.57 MB → 4.49 MB (−80 KB)**.

## Manual E2E verification (simulator, Prod)
Ran a real placement end to end on an iOS 18.5 simulator: Example app → Environment **Prod** → Mobile-team test tag → Initialize → Cart → Page Identifier `MSDKOverlayLayout` → **Pay 40 USD** → `selectPlacements`. Temporary trace logs (reverted, not in this PR) confirmed the new path on a live ~200 KB offers response:

```
adapter: RenderExperience(single-pass) encoded 200271 bytes;
   responseOptionsMap=true  catalogItems=false  camelToken=true  snakeLeak=false
parse:   parseExperience OK  pageModel=non-nil  pageId=698b87c2-…
```

- The single-pass encoder produced the renderer's **camelCase** contract (`responseOptionsMap`, `token`) with **no snake_case leakage** (`session_token` absent).
- `RoktUX.parseExperience` decoded it into a **non-nil page model**.
- The overlay **rendered correctly** — offer copy plus the positive/negative response options as "Redeem now" / "Decline", confirming the `responseOptionsMap` bucketing survives the full fetch → decode → encode → parse → render path.

> Note: commits are unsigned — the local 1Password commit-signing agent was unavailable in the automation environment. Re-sign before merge if branch protection requires it.

### DIFF
--- a/Sources/Rokt_Widget/Models/RenderExperience.swift
+++ b/Sources/Rokt_Widget/Models/RenderExperience.swift
@@ -1,278 +1,231 @@
-// periphery:ignore:all - offers render models; fields are encode-only
+// periphery:ignore:all - offers render encoder; encode-only
 import Foundation
 
-// MARK: - Renderer experience contract (camelCase)
-
-/// The experience JSON the renderer decodes (`RoktUXExperienceResponse`), mapped
-/// from the offers ``SelectResponse``. The pre-serialized DCUI layout-schema
-/// strings pass through verbatim; the offer/creative subtree is re-homed from the
-/// response's snake_case to the renderer's camelCase contract.
+/// Encodes a ``SelectResponse`` into the renderer's experience JSON
+/// (`RoktUXExperienceResponse`, camelCase) in a single pass, with no parallel
+/// model tree. The pre-serialized DCUI layout-schema strings pass through
+/// verbatim; the offer/creative subtree is re-homed from the response's
+/// snake_case to the renderer's camelCase contract.
+///
+/// Field presence matches the renderer's decode contract: required keys are
+/// always emitted (defaulting to `""` / `[]` / `{}`), optional keys are omitted
+/// when absent. The renderer decodes with a plain `JSONDecoder` (no key
+/// strategy), so the keys written here must match its property names verbatim.
 internal struct RenderExperience: Encodable {
-    let sessionId: String
-    let token: String
-    let page: RenderPage
-    let placementContext: RenderPlacementContext
-    let placements: [RenderPlacement]
-    let plugins: [RenderPluginContainer]?
+    private let response: SelectResponse
 
     init(_ response: SelectResponse) {
-        sessionId = response.sessionId
-        token = response.sessionToken.token
-        page = RenderPage(pageId: response.pageContext?.pageId)
-        placementContext = RenderPlacementContext(response)
+        self.response = response
+    }
+
+    /// Dynamic string coding key so the whole render contract is expressed with
+    /// literal key names, without a coding-key enum per nested shape.
+    private struct Key: CodingKey {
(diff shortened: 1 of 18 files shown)


Sources/Rokt_Widget/Models/RenderExperience.swift
```diff
@@ -1,278 +1,231 @@
...
-    let token: String
 
-    init(_ response: SelectResponse) {
-        // Not in the response; required by the decoder but unused when rendering.
-        roktTagId = ""
-        pageInstanceGuid = response.pageContext?.pageInstanceGuid ?? response.pageInstanceGuid
-        token = response.pageContext?.token ?? response.sessionToken.token
+    /// Dynamic string coding key so the whole render contract is expressed with
+    /// literal key names, without a coding-key enum per nested shape.
+    private struct Key: CodingKey {
```
denischilik, line 23: (optional to check) we may try to change this struct to 

private struct Key: CodingKey, ExpressibleByStringLiteral {
    let stringValue: String
    var intValue: Int? { nil }
    init(_ stringValue: String) { self.stringValue = stringValue }
    init?(stringValue: String) { self.stringValue = stringValue }
    init?(intValue: Int) { nil }
    init(stringLiteral value: String) { self.stringValue = value }
}

then you can write keys like this

try container.encode(response.sessionId, forKey: "sessionId")
try container.encodeIfPresent(layout?.name, forKey: "name")

denischilik (approved):

Sources/Rokt_Widget/Models/RenderExperience.swift
nickolas-dimitrakas: Will keep this in mind when we come back to this in the follow up.

PR #267, opened 2026-07-28
fix(fonts): always settle font batch and recover unreadable cached fonts

nickolas-dimitrakas:
## Background

Since 5.2.6 the font cache lives in `Library/Caches` (#235). That is the correct
location for re-downloadable data, but it changes an assumption the font code was
built on: under `Documents` a font was fetched once per install and then stayed
put, so cache misses were effectively impossible after first launch. Under
`Caches` the OS may purge entries at any time, without notice, and a purge can
leave a partially written file behind. Paths that were previously almost never
taken are now routine, and two of them do not degrade gracefully.

**Completion could never fire.** `downloadFonts` picked the download that owned
the completion handler by comparing a running counter against `fonts.count` at
the moment each download was issued. If any font needed downloading but the last
element of the array did not — because it was already cached, or resolved to a
system font — then no download ever carried the flag and `onFontDownloadComplete`
was never called. A mid-list font that could not resolve a file URL also
`return`ed out of the entire batch without settling. Both leave the caller
waiting forever.

**An unreadable cached font never recovered.** `registerFont` reported read and
registration failures via diagnostics but left the offending cache entry on disk.
Because `isDownloadingFontRequired` treats any present file as usable, the same
unreadable bytes were re-read and re-rejected on every render, pinning the layout
to the fallback font for the life of the install. Only the 7-day expiry could
clear it, and only when the temporary-font-cache flag is enabled.

## What Has Changed

- `downloadFonts` now counts fonts as they settle and fires completion exactly
  once when the last one lands, via a small `FontDownloadBatch` helper. Every
  outcome settles: downloaded, loaded from cache, skipped as a system font,
  failed to resolve a file URL, or failed to download. Ordering no longer
  matters, and the mid-list early `return` is gone.
- `RoktNetWorkAPI.downloadFont` drops the `isLastFont` in/out parameter, which
  existed only to route completion. Its callback now fires on every terminal
  path, including the previously uncovered case where a response carried neither
  an error nor a local file URL. `finishedDownloadingFonts` is posted once by
  `FontManager`, which owns the batch, instead of by the network layer.
- Read and registration failures now invalidate the cache entry (file plus
  metadata) so the next pass re-fetches the font instead of resolving to the
  fallback font indefinitely. Recovery is capped per font per process
  (`maxFontRecoveryAttempts`), uses a constant backoff, and runs asynchronously,
  so a permanently bad asset cannot loop and a re-fetch never delays a render.
- A file that fails immediately after being downloaded is treated as a bad asset
  rather than a bad cache: it is not re-fetched, but it is still removed so a
  poisoned file is not left for the next launch to read.
- Recovery budget resets on `initialize`, alongside the existing
  `isInitFailedForFont` reset.
- Removed the unreachable second `isSystemFont` branch in `downloadFonts` and its
  now-unused `[FFL002]` code. The preceding `guard` already excludes system fonts.

Behaviour deliberately left alone: font failures still only mark
`isInitFailedForFont` for diagnostics and never un-initialise the SDK.

## How Has This Been Tested

Four tests added to `TestFontManager`:

- `test_downloadFonts_callsCompletion_whenLastFontIsAlreadyCached`
- `test_downloadFonts_callsCompletion_whenLastFontIsSystemFont`
- `test_registerFont_withUnregisterableCachedFont_invalidatesCacheForReDownload`
- `test_registerFont_withUnregisterableCachedFont_stopsRecoveringAtAttemptLimit`

The two completion tests were first run against unmodified `main` in a separate
worktree to confirm they reproduce the bug rather than merely describing the new
behaviour. Both timed out waiting for `onFontDownloadComplete`; both pass on this
branch.

Full package suite on iPhone 16 Pro (iOS 18.5): **653 tests, 0 failures**,
2 skipped. `trunk check` clean on all four changed files.

## Notes

The completion bug is only observable when a batch mixes fonts that need
downloading with fonts that do not, which is the normal steady state once a
cache is partially populated or partially purged — so it is far more reachable
after the move to `Caches` than it was before, even though the logic itself
predates that change.

## Font downloads are no longer bound by the API client timeout

Added after production data showed this is the dominant failure, not the cache bug. Over two
days `com.klarna.app` on 5.2.6 produced ~559k `[FONT]` errors with `NSURLErrorDomain Code=-1001`
("The request timed out"), against ~47k registration failures and ~9k read failures.

`updateTimeout` applied the client timeout to `timeoutIntervalForResource`, which bounds an
*entire* transfer rather than idle time. At the default 9 seconds, a font needs sustained
multi-Mbps throughput to survive regardless of how healthy the connection is. Two of this
account's faces are 5.68 MB and 5.78 MB, so they need ~5 Mbps each; because the download loop
starts every font at once on one session, they also starve the small ones into the same
deadline — ~108k of the timeouts are on two ~138 KB files.

The 30 second `defaultFontTimeout` never helped: it lands on `URLRequest.timeoutInterval`,
which is an idle timeout, and a per-request value cannot lift the session's resource ceiling.

Downloads now run on `downloadSession` — already constructed but never used for any request —
with the idle timeout doing the real work and a large resource ceiling only to stop a
pathological trickle. Its configuration is copied from the caller's so injected protocol
classes still apply, which keeps the mocked download tests intercepting.

Guarded by three tests, including one that cancels the API session to prove downloads are not
issued on it. Reverting the session swap fails them.

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
+++ b/Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
@@ -70,9 +70,13 @@ internal final class RoktHTTPClient: HTTPClientAdapter {
     private(set) var downloadSession: URLSession = .shared
     private(set) var encoders: [RoktHTTPParameterEncoder] = []
 
-    /// Ceiling on an entire download, as opposed to the per-request budget, which only has to
-    /// cover idle time. Deliberately far larger than the API client timeout: it exists to stop
-    /// a pathological trickle running forever, not to bound a healthy transfer.
+    /// How long a download may sit without receiving data before it is abandoned. This, rather
+    /// than a cap on the total transfer, is what should fail a download: a large file on a slow
+    /// connection is still making progress, whereas one receiving nothing is not.
+    static let downloadIdleTimeout: TimeInterval = 30
+
+    /// Ceiling on an entire download. Deliberately far larger than the API client timeout: it
+    /// exists to stop a pathological trickle running forever, not to bound a healthy transfer.
     static let downloadResourceTimeout: TimeInterval = 300
 
     init(
@@ -90,13 +94,17 @@ internal final class RoktHTTPClient: HTTPClientAdapter {
     /// `timeoutIntervalForResource` caps a whole transfer rather than idle time, so applying
     /// the API client timeout to it cancels any download that cannot finish inside that
     /// budget no matter how healthy the connection is. Font files are orders of magnitude
-    /// larger than API payloads, so downloads get their own session and are bounded by the
-    /// per-request idle timeout instead. Derived from the caller's configuration so injected
-    /// protocol classes still apply.
+    /// larger than API payloads, so downloads get their own session bounded by idle time.
+    ///
+    /// Both intervals are set explicitly rather than inherited: the incoming configuration is
+    /// sized for API calls, so leaving either in place would silently reimpose an API-shaped
+    /// budget on file transfers. Everything else is copied so injected protocol classes and
+    /// cache policy still apply.
     private static func downloadConfiguration(
         from configuration: URLSessionConfiguration
     ) -> URLSessionConfiguration {
         let downloadConfiguration = (configuration.copy() as? URLSessionConfiguration) ?? configuration
+        downloadConfiguration.timeoutIntervalForRequest = downloadIdleTimeout
         downloadConfiguration.timeoutIntervalForResource = downloadResourceTimeout
 
         return downloadConfiguration
--- a/Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
+++ b/Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
@@ -7,6 +7,8 @@ enum RoktDownloadOptions: CaseIterable {
 
 // Protocol to hide implementation details
 protocol HTTPClientAdapter {
+    /// - Parameter timeout: seconds. Callers holding a millisecond-based config value, such as the
+    ///   client timeout from init, must convert before calling.
     func updateTimeout(timeout: Double)
 
     @discardableResult
@@ -73,11 +75,11 @@ internal final class RoktHTTPClient: HTTPClientAdapter {
     /// How long a download may sit without receiving data before it is abandoned. This, rather
     /// than a cap on the total transfer, is what should fail a download: a large file on a slow
     /// connection is still making progress, whereas one receiving nothing is not.
-    static let downloadIdleTimeout: TimeInterval = 30
+    static let downloadIdleTimeoutSeconds: TimeInterval = 30
 
     /// Ceiling on an entire download. Deliberately far larger than the API client timeout: it
     /// exists to stop a pathological trickle running forever, not to bound a healthy transfer.
-    static let downloadResourceTimeout: TimeInterval = 300
+    static let downloadResourceTimeoutSeconds: TimeInterval = 300
 
     init(
         sessionConfiguration: URLSessionConfiguration = .default,
@@ -104,14 +106,16 @@ internal final class RoktHTTPClient: HTTPClientAdapter {
         from configuration: URLSessionConfiguration
     ) -> URLSessionConfiguration {
         let downloadConfiguration = (configuration.copy() as? URLSessionConfiguration) ?? configuration
-        downloadConfiguration.timeoutIntervalForRequest = downloadIdleTimeout
-        downloadConfiguration.timeoutIntervalForResource = downloadResourceTimeout
+        downloadConfiguration.timeoutIntervalForRequest = downloadIdleTimeoutSeconds
+        downloadConfiguration.timeoutIntervalForResource = downloadResourceTimeoutSeconds
 
         return downloadConfiguration
     }
 
     /// Applies to API requests only. Downloads keep their own budget, since the client timeout
     /// is sized for small JSON payloads and would cancel large files mid-transfer.
+    ///
+    /// - Parameter timeout: seconds.
     func updateTimeout(timeout: Double) {
         let currentConfiguration = session.configuration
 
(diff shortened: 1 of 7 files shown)


Sources/Rokt_Widget/Networking/RoktNetworkAPI.swift
```diff
@@ -61,40 +60,35 @@ internal class RoktNetWorkAPI {
...
+
+                return
+            } else {
+                // Best-effort: mark for diagnostics, don't un-initialise.
+                Rokt.shared.roktImplementation.isInitFailedForFont = true
+                let callstack = "\(fontErrorMessage) \(font.url), " +
+                    "error: \(String(describing: downloadResponse.downloadError.debugDescription))"
+
+                RoktAPIHelper.sendDiagnostics(message: fontDiagnosticCode, callStack: callstack)
+                RoktLogger.shared.verbose(callstack)
```
BrandonStalnaker: Settling on the previously uncovered nil-error/nil-URL path is correct. Folding it into the same diagnostic that says Error downloading font: with Optional.none is not — you'll get confusing [FONT] rows and a sticky isInitFailedForFont for a malformed response, not a download failure. Split "terminal download error" from "response missing both error and file URL".

Sources/Rokt_Widget/Util/FontManager.swift
```diff
@@ -271,6 +268,68 @@ internal class FontManager {
...
+    /// Removes the cached file and its metadata so `isDownloadingFontRequired` reports
+    /// the font as missing on the next pass.
+    internal static func invalidateCachedFont(_ font: FontModel, completion: (() -> Void)? = nil) {
+        if let fileUrl = getFileUrl(name: font.postScriptName ?? font.name),
+           FileManager.default.fileExists(atPath: fileUrl.path) {
+            try? FileManager.default.removeItem(at: fileUrl)
+        }
+
+        FontRepository.removeFontDetail(key: font.url) { completion?() }
+    }
```
BrandonStalnaker, line 338: I think this deletes the file + detail JSON, but not the entry in RoktFontDownloadedUrl. isDownloadingFontRequired only needs the detail, so re-download still works, but removeUnusedFonts / removeFont short-circuit on missing detail and leave orphans. You should either also call removeFontUrl, or reuse removeFont.

BrandonStalnaker (changes requested):
Mainly want diagnostic-path split around downloadFont else-branch before merge so we don't invent a new noisy [FONT] shape while the existing one is already getting hit so much.

Sources/Rokt_Widget/Networking/RoktNetworkAPI.swift
nickolas-dimitrakas: Agreed — self-inflicted. Flattening the retry guard into one condition widened the trailing `else`, so terminal download errors and "neither error nor file" responses ended up sharing a message, hence the `Optional.none`.

Split in 7284318. The new row reads `Font download response missing both error and file: <url>, status: <code|none>`.

I kept the terminal message byte-for-byte so existing `[FONT]` queries keep matching, with a test pinning that shape (verified it stays green when the split is reverted).

Extracted `handleFontDownloadResponse` to test it — `RoktHTTPClient` is `final` and never emits that shape itself, which is why the bad output went unnoticed.

One call for you: I left `isInitFailedForFont = true` on the malformed branch, since the font still didn't land and the flag is diagnostic-only post-5.3.0. Say the word if you'd rather it not set at all.

Sources/Rokt_Widget/Util/FontManager.swift
nickolas-dimitrakas: Right on both halves. Re-download does still work since `isDownloadingFontRequired` only reads the detail — but `removeFont` guards on the detail before it reaches `removeFontUrl`, so once the detail is gone the URL entry is permanently unreachable and `RoktFontDownloadedUrl` grows unbounded across evictions.

Fixed in 7284318 — removes both, chained so `completion` still means both writes landed:

```swift
FontRepository.removeFontDetail(key: font.url) {
    FontRepository.removeFontUrl(key: font.url) { completion?() }
}
```

Didn't reuse `removeFont` because it derives the filename from the stored detail — the record most likely missing or stale in the eviction cases this path exists for — and it has no completion handler.

Covered by `test_invalidateCachedFont_alsoRemovesTheDownloadedUrlEntry`, confirmed failing against the detail-only version.

denischilik (commented):
Could you pleace use Actors instead NSLock

BrandonStalnaker (dismissed):

Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
```diff
@@ -70,16 +70,48 @@ internal final class RoktHTTPClient: HTTPClientAdapter {
     private(set) var downloadSession: URLSession = .shared
     private(set) var encoders: [RoktHTTPParameterEncoder] = []
 
+    /// How long a download may sit without receiving data before it is abandoned. This, rather
+    /// than a cap on the total transfer, is what should fail a download: a large file on a slow
+    /// connection is still making progress, whereas one receiving nothing is not.
+    static let downloadIdleTimeout: TimeInterval = 30
```
thomson-t: Would be nice to add explicit Millis/Sec

nickolas-dimitrakas: Good call — and worse than it looked from the diff, since the same file that seeds these has `defaultTimeout = 9000` (ms) sitting directly above `defaultFontTimeout = 30` (s, unit only in a trailing comment).

Renamed on this path in 46f536d:

- `downloadIdleTimeout` → `downloadIdleTimeoutSeconds`
- `downloadResourceTimeout` → `downloadResourceTimeoutSeconds`
- `defaultFontTimeout` → `defaultFontTimeoutSeconds` (dropped the `// second` comment)
- `defaultTimeout` → `defaultTimeoutMilliseconds`

Also documented `updateTimeout(timeout:)` as seconds. Its only caller is `NetworkingHelper.updateTimeout(timeout: clientTimeoutMilliseconds/1000)`, so the `/1000` was the only place the unit was stated. I left the parameter label alone rather than churn the protocol and its 7 conformances in a hotfix — happy to rename it to `timeoutSeconds` in a follow-up if you'd prefer it enforced at the call site.

BrandonStalnaker (approved):

PR #268, opened 2026-07-28
refactor(fonts): isolate font download batch state in an actor

nickolas-dimitrakas:
Stacked on #267 — targets `fix/font-cache-recovery`, since both types this touches are introduced there and do not exist on `main`. Review/merge #267 first; this will retarget to `main` automatically.

Addresses @denis-rokt's review comment on #267 asking for actors instead of `NSLock`.

## What changed

**`FontDownloadBatch` is now an actor.** Its counter is shared across download callbacks that arrive on arbitrary threads, so it is a good fit: actor isolation lets the compiler enforce the exclusion rather than a hand-held lock. `settle()` now returns `Bool` — `true` only for the call that settles the last outstanding font — so the completion runs off the actor and cannot fire twice.

Completion is dispatched to the main queue. The download callbacks already deliver there (`RoktHTTPClient` completes on `.main`), so observers of `finishedDownloadingFonts` keep their existing thread, and the `queue.sync(flags: .barrier)` inside `getExistingFontsByPostScriptName()` stays off the cooperative thread pool.

**The recovery budget stays synchronous, with its lock encapsulated.** It is read and written from `registerFont`, which `reRegisterFonts` calls on the render path. Making it an actor forces `registerFont` to be `async`, which propagates to `reRegisterFonts` and turns the execute continuation at `RoktInternalImplementation.swift:1087` — the one that drives the cache lookup, `isExecuting`, and presentation — into a deferred hop. That is a render-path behaviour change unrelated to the bug #267 fixes, so instead the state moved into a dedicated `FontRecoveryLedger` that owns its lock. No call site handles `lock()`/`unlock()` anymore.

`@Sendable` was threaded through `downloadFonts` to keep the build warning-free; that surfaced a real data race in `test_downloadFonts_should_call_onFontDownloadComplete_only_once...`, where an unguarded `var callbackCount` was mutated from a concurrent callback. It now uses the existing thread-safe `Counter`.

## No behaviour change

No test changes were needed beyond that race fix — the batch's observable contract (fires exactly once, regardless of font order) is unchanged and still covered by the existing tests.

## Test plan

- [x] `xcodebuild build` — no new warnings
- [x] Full `Rokt_WidgetTests` suite: 660 tests, 0 failures, 2 skipped
- [x] `TestFontManager`: 35 tests, 0 failures
- [x] `trunk check` clean on all changed files

### DIFF
--- a/Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
+++ b/Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
@@ -70,16 +70,40 @@ internal final class RoktHTTPClient: HTTPClientAdapter {
     private(set) var downloadSession: URLSession = .shared
     private(set) var encoders: [RoktHTTPParameterEncoder] = []
 
+    /// Ceiling on an entire download, as opposed to the per-request budget, which only has to
+    /// cover idle time. Deliberately far larger than the API client timeout: it exists to stop
+    /// a pathological trickle running forever, not to bound a healthy transfer.
+    static let downloadResourceTimeout: TimeInterval = 300
+
     init(
         sessionConfiguration: URLSessionConfiguration = .default,
         encoders: [RoktHTTPParameterEncoder] = [RoktHTTPURLEncoder(), RoktHTTPBodyEncoder()]
     ) {
         self.session = URLSession(configuration: sessionConfiguration)
-        self.downloadSession = URLSession(configuration: sessionConfiguration)
+        self.downloadSession = URLSession(
+            configuration: Self.downloadConfiguration(from: sessionConfiguration)
+        )
 
         self.encoders = encoders
     }
 
+    /// `timeoutIntervalForResource` caps a whole transfer rather than idle time, so applying
+    /// the API client timeout to it cancels any download that cannot finish inside that
+    /// budget no matter how healthy the connection is. Font files are orders of magnitude
+    /// larger than API payloads, so downloads get their own session and are bounded by the
+    /// per-request idle timeout instead. Derived from the caller's configuration so injected
+    /// protocol classes still apply.
+    private static func downloadConfiguration(
+        from configuration: URLSessionConfiguration
+    ) -> URLSessionConfiguration {
+        let downloadConfiguration = (configuration.copy() as? URLSessionConfiguration) ?? configuration
+        downloadConfiguration.timeoutIntervalForResource = downloadResourceTimeout
+
+        return downloadConfiguration
+    }
+
+    /// Applies to API requests only. Downloads keep their own budget, since the client timeout
+    /// is sized for small JSON payloads and would cancel large files mid-transfer.
     func updateTimeout(timeout: Double) {
         let currentConfiguration = session.configuration
 
@@ -272,7 +296,7 @@ extension RoktHTTPClient {
             return
         }
 
-        session.downloadTask(with: downloadRequest) { [weak self] temporaryURL, downloadResponse, downloadError in
+        downloadSession.downloadTask(with: downloadRequest) { [weak self] temporaryURL, downloadResponse, downloadError in
             guard let self else { return }
 
             if let downloadError {
--- a/Sources/Rokt_Widget/Util/FontManager.swift
+++ b/Sources/Rokt_Widget/Util/FontManager.swift
@@ -319,15 +319,22 @@ internal class FontManager {
         }
     }
 
-    /// Removes the cached file and its metadata so `isDownloadingFontRequired` reports
-    /// the font as missing on the next pass.
+    /// Removes the cached file and both metadata records so `isDownloadingFontRequired`
+    /// reports the font as missing on the next pass.
+    ///
+    /// Dropping the detail alone would be enough to trigger a re-download, but it would
+    /// strand the URL entry: `removeFont` reads the detail first and returns early when it
+    /// is absent, so `removeUnusedFonts` could never clean the entry up afterwards. A
+    /// successful re-download re-adds both records.
     internal static func invalidateCachedFont(_ font: FontModel, completion: (() -> Void)? = nil) {
         if let fileUrl = getFileUrl(name: font.postScriptName ?? font.name),
            FileManager.default.fileExists(atPath: fileUrl.path) {
             try? FileManager.default.removeItem(at: fileUrl)
         }
 
-        FontRepository.removeFontDetail(key: font.url) { completion?() }
+        FontRepository.removeFontDetail(key: font.url) {
+            FontRepository.removeFontUrl(key: font.url) { completion?() }
+        }
     }
 
     #if DEBUG
--- a/Sources/Rokt_Widget/Util/FontManager.swift
+++ b/Sources/Rokt_Widget/Util/FontManager.swift
@@ -9,7 +9,6 @@ internal class FontManager {
     static let downloadingFonts = "downloadingFonts"
     private static let fullFontLogDiagnosticCode = "[FULLFONTLOGS]"
     private static let fullFontLogCode1 = "[FFL001]"
-    private static let fullFontLogCode2 = "[FFL002]"
     private static let fullFontLogCode5 = "[FFL005]"
     private static let fullFontLogCode6 = "[FFL006]"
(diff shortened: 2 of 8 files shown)


Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
```diff
@@ -72,14 +70,10 @@ internal final class RoktHTTPClient: HTTPClientAdapter {
...
-    /// connection is still making progress, whereas one receiving nothing is not.
-    static let downloadIdleTimeoutSeconds: TimeInterval = 30
-
-    /// Ceiling on an entire download. Deliberately far larger than the API client timeout: it
-    /// exists to stop a pathological trickle running forever, not to bound a healthy transfer.
-    static let downloadResourceTimeoutSeconds: TimeInterval = 300
+    /// Ceiling on an entire download, as opposed to the per-request budget, which only has to
+    /// cover idle time. Deliberately far larger than the API client timeout: it exists to stop
+    /// a pathological trickle running forever, not to bound a healthy transfer.
+    static let downloadResourceTimeout: TimeInterval = 300
```
thomson-t, line 76: Looks like this is 300 seconds(5 minutes).
How about `downloadResourceTimeoutSeconds` to make it clear?

Sources/Rokt_Widget/Util/FontManager.swift
```diff
@@ -62,78 +61,70 @@ internal class FontManager {
         completionHandler?()
     }
 
-    class func downloadFonts(_ fonts: [FontModel], _ onFontDownloadComplete: @escaping () -> Void) {
+    class func downloadFonts(_ fonts: [FontModel], _ onFontDownloadComplete: @escaping @Sendable () -> Void) {
         getExistingFontsByPostScriptName()
```
roktfleet: **Stale branch in conflict with main; a resolution toward the branch would revert main's later font changes** (blocking) — GitHub reports the pull request unmergeable (mergeable_state dirty): the base is now main (auto-retargeted when #267 merged on 2026-07-29T14:23Z) and the branch still carries #267's pre-squash commits. main's downloadFonts calls FontRepository.migrateLegacyFontStorageIfNeeded() on the line after this one and gates downloads on isFontDownloadBlockedByDiskPressure() further down (read from main this round); neither name appears anywhere in this head's diff, so a conflict resolved toward the branch would silently revert them. The fleet's lesson 36, a stale branch resolved toward the branch reverts main: rebase onto current main dropping #267's commits and confirm the file list shows only the actor, ledger, @Sendable and test-counter changes.

Sources/Rokt_Widget/Util/FontManager.swift
```diff
@@ -62,78 +61,70 @@ internal class FontManager {
...
+    }
 
-        if !fontDownloadInProgress {
-            onFontDownloadComplete()
+    /// Settles one font against `batch` and, for whichever call completes it, runs
+    /// `complete` on the main queue. Download callbacks already land there, so keeping the
+    /// batch completion on main leaves `finishedDownloadingFonts` observers where they were
+    /// and keeps the barrier in `getExistingFontsByPostScriptName` off the cooperative pool.
+    private static func finish(_ batch: FontDownloadBatch, then complete: @escaping @Sendable () -> Void) {
+        Task {
```
roktfleet: **"No behaviour change" does not hold for the non-download path: batch completion moves from synchronous to two asynchronous hops** (should-fix) — On main FontDownloadBatch is a lock-guarded class whose settle() calls onComplete() directly, on whichever thread settles last (its own doc comment says so), so a batch of only cached or system fonts calls onFontDownloadComplete() and posts finishedDownloadingFonts before downloadFonts returns. With the actor every completion is deferred through this Task, an actor hop and DispatchQueue.main.async. The description does say completion is dispatched to the main queue, but its claims that observers keep their existing thread and that there is no behaviour change only hold when the last font to settle was a download (RoktHTTPClient's completionQueue defaults to .main); for cached and system fonts the thread and the timing both change, and the deferred hop is the shape the body says it avoided for registerFont. The fleet's lesson 30, the description is the defect: say so in the body and confirm the init continuation tolerates a callback that arrives after downloadFonts has returned.

Sources/Rokt_Widget/Networking/HTTPClient/RoktHTTPClient.swift
```diff
@@ -272,7 +296,7 @@ extension RoktHTTPClient {
             return
         }
 
-        session.downloadTask(with: downloadRequest) { [weak self] temporaryURL, downloadResponse, downloadError in
+        downloadSession.downloadTask(with: downloadRequest) { [weak self] temporaryURL, downloadResponse, downloadError in
             guard let self else { return }
```
roktfleet: **A deallocated client drops the download completion, so a font can settle zero times** (question) — If the RoktHTTPClient is released while a font download is in flight (deinit calls downloadSession.finishTasksAndInvalidate(), so the task still completes; the test mock's stubFontFileUrl replaces NetworkingHelper.shared.httpClient), this guard returns without invoking the completion, the font never settles, and the batch and any init path waiting on it never complete. This is a context line, not this PR's change, but the actor's exactly-once contract rests on it; the fleet's lesson 42, silent lane death behind green. sendDownloadResultCallback captures only completionQueue and completionHandler, not self, so the else branch could report a cancellation error instead of returning silently. Is the shared client ever replaced in production?

roktfleet (commented):
[SDK] fleet review — read 2026-09-08T09:02Z on the current head. Machine-written by the Rokt fleet's SDK agent; one comment review, no approval or request-changes, and nothing is pushed or merged here by the fleet.

**What this change does (as read from the diff)**
`FontDownloadBatch` becomes an actor whose `settle()` returns `true` only for the completing call; `FontManager.finish` runs that completion via a `Task` and `DispatchQueue.main.async`. The recovery attempt count moves into a `FontRecoveryLedger` owning its `NSLock`; `@Sendable` is threaded through `downloadFonts`; one test counter becomes `Counter`. The diff also carries #267's changes, merged to `main` on 29 July.

**Findings**
1. **Blocking** — Stale branch, in conflict with `main` (GitHub: unmergeable). `main`'s `downloadFonts` now calls `FontRepository.migrateLegacyFontStorageIfNeeded()` and gates on `isFontDownloadBlockedByDiskPressure()`; this head has neither, so a conflict resolved toward the branch would silently revert them (the fleet's lesson 36, a stale branch resolved toward the branch reverts main). Rebase onto current `main`, dropping #267's commits; confirm the file list is only this PR's change.
2. **Should-fix** — "No behaviour change" does not hold for the non-download path. On `main` the completing `settle()` runs the completion synchronously, so an all-cached or all-system batch completes before `downloadFonts` returns; here every completion goes through a `Task` and a `main.async` hop. The body says completion is dispatched to main, but "observers keep their existing thread" only holds when the last font was a download. Say so; confirm the init continuation tolerates the later callback (lesson 30, the description is the defect).
3. **Question** — `guard let self else { return }` in `downloadFile`'s task callback drops the completion, so a font never settles if the client is deallocated mid-download (`deinit` calls `finishTasksAndInvalidate`; the test mock swaps `NetworkingHelper.shared.httpClient`). Pre-existing, but the exactly-once contract rests on it; `sendDownloadResultCallback` captures no `self`, so the `else` branch could report a cancellation (lesson 42, silent lane death behind green).

**Bar check**
Checks on this head all SUCCESS: SPM Unit Tests, UI Tests, Trunk Check, Podspec Lint, Periphery Scan, Consumer Test (PactSwift), Detect Contract Changes, SDK Size Report, codecov, paved-roads-agent, pr-security-agent (Publish to Broker and Can I Deploy skipped); all ran 28 July, before #267 merged. No Claude code review on this head. Description has What and Why but still names `fix/font-cache-recovery` as its target. 895 changed lines, 8 files, none generated — over the 400-line bar. One open human thread (thomson-t, `downloadResourceTimeout` naming). No open bot findings. Draft, unmergeable. Public repository: SDK leads merge.

**To reach ready**
- Rebase onto current `main`; verify the file list is only this PR's change.
- Update the description: targets `main`; note the deferred completion for cached and system fonts.
- Answer thomson-t's thread, then mark ready for the SDK leads.

PR #271, opened 2026-07-30
fix: add contract check to Release Draft and rename PR gate

denischilik:
## Background

Keep the blocking production contract check on PRs (merge gate), rename it so it is clear this is not a deploy step, and add the same check to Release – Draft so a release cannot start against an unverified main contract.

## What Has Changed

- PR job renamed to `Contract verified vs production` (still blocking via `can-i-deploy --to-environment production`)
- Release – Draft runs the same blocking check for `--latest main` before opening the release PR
- Clarifies intended dual gate: merge requires prod-verified contract; release re-checks latest main

## How Has This Been Tested

- `trunk check` clean on both workflow files
- Pact Consumer CI on this PR exercises the PR gate
- Draft gate will be exercised on the next manual Release – Draft run after merge

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works.


.github/workflows/release-draft.yml
```diff
@@ -21,7 +21,40 @@ permissions:
   pull-requests: write
```
thomson-t: Please address the comments from Zizmor scan

denischilik: Fixed the excessive-permissions finding: workflow level is now `contents: read`, and `contents: write` / `pull-requests: write` are scoped to the job that opens the release PR. The new `can-i-deploy` job no longer inherits write.

The remaining annotations on this file (unpinned `generate-changelog@main`, checkout without `persist-credentials: false`) predate this PR — happy to fix them here too if you'd prefer.

.github/workflows/release-draft.yml
```diff
@@ -17,12 +17,41 @@ concurrency:
...
+              --to-environment production
+
   prepare-release:
+    needs: [can-i-deploy]
     runs-on: macos-latest
+    permissions:
+      contents: write
+      pull-requests: write
     steps:
       - name: Checkout repository
```
thomson-t, line 56: While we are updating the permissions, can we address the other security comments as well? `persist-credentials: false` 
and pin the `ROKT/rokt-workflows/actions/generate-changelog` to a commit hash so that the security check is green.

denischilik: Done — checkout now sets `persist-credentials: false`, and `generate-changelog` is pinned to `1859338d7c2c2b873233505c751acbf09fb218d5` (same hash as Android).

thomson-t (approved):

PR #272, opened 2026-07-30
ci: Deepen coverage for background flush and pending event store

BrandonStalnaker:
## Background

Deepens unit-test coverage for the interaction between background event flush and the persisted pending-event store.

When the app backgrounds, buffered transaction events are flushed into the send pipeline. If that send fails (for example offline or recoverable gateway errors), the batch is written to the on-disk pending store for replay on the next successful init. Review feedback on 
[#251](https://github.com/ROKT/rokt-sdk-ios/pull/251) called out that this can rewrite the pending store from a background context when paired with 
[#250](https://github.com/ROKT/rokt-sdk-ios/pull/250), and asked for further investigation beyond the original unit tests.

This PR adds concurrency and composed-path tests for that handoff. No production code changes. No new dependencies.

## What Has Changed

- TestTxnPendingEventStore: concurrent persist (file stays decodable and within cap), concurrent persist+drain thread-safety, and persist-after-drain rewriting a fresh store.
- TestTxnEventService: exhausted transport / offline failure persists the batch for replay (background-relevant failure mode).
- TestEventQueue: flush cancels the debounce timer (no double delivery); flush concurrent with enqueue still delivers once.
- TestBackgroundFlushPendingStore (new): composed path — didEnterBackground → flush → failed send → persist → cold drain; plus cross-“process” persist then drain shape check.

## How Has This Been Tested

Ran the affected suites locally via the Rokt-Widget scheme on an iPhone 16 Pro simulator:
Result: 22 tests executed, 0 failures (1 existing Data Protection assertion skipped on simulator).

## Checklist

### DIFF
--- a/Tests/Rokt_WidgetTests/Shared/DataAccess/TestTxnPendingEventStore.swift
+++ b/Tests/Rokt_WidgetTests/Shared/DataAccess/TestTxnPendingEventStore.swift
@@ -114,4 +114,83 @@ final class TestTxnPendingEventStore: XCTestCase {
         XCTAssertEqual(drained.count, TxnPendingEventStore.maxBatches)
         XCTAssertFalse(drained.contains { $0.first?.instanceId == "stale" })
     }
+
+    // MARK: - Background rewrite / concurrency (#250 flush → #251 persist)
+
+    /// Multiple failed sends (e.g. background flush draining several Tasks) rewrite the same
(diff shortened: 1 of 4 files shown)


jamesnrokt (approved):
Thanks for covering this. Do we also need any functionality to send diagnostics events to notify us of an issue if something fails to write?

PR #275, opened 2026-08-03
feat: Update to Session Token

BrandonStalnaker:
## Background

Hybrid native + WebView integrations need to share more than a session id. Offers and events authorize with a short-lived session token (Bearer JWT) plus expiry, but the public setSessionId / getSessionId APIs only updated the legacy diagnostics session id and did not seed that token for the next offers/events call.

This PR adds additive APIs so partners can hand off the full session (id + token + expiry) and keep continuity across integrations. Existing session-id-only APIs stay source/binary compatible and are marked deprecated.

## What Has Changed

- Added public @objc RoktSession (sessionId, sessionToken, expiresAt in Unix epoch ms).
- Added Rokt.setSession(_:) and Rokt.getSession().
- setSession synchronously persists into the txn session store used by offers/events (so the next call can send Authorization: Bearer …), and mirrors the session id into the legacy session manager.
- getSession reads that store with the same tag-id / expiry rules; returns nil when uninitialized, missing, or expired.
- Deprecated setSessionId(sessionId:) and getSessionId() (behavior unchanged: legacy session id only).
- Extracted shared sync persist/load helpers used by TxnSessionManager and the public handoff path.
- Extended unit tests for round-trip, empty/blank no-ops, init gate, expiry clear, Bearer restore, and deprecated BC.

## How Has This Been Tested

- trunk check on changed files
- Unit tests: xcodebuild test -scheme Rokt-Widget for TestRokt and TestTxnSessionManager (all passed)
- xcodebuild build for rokt-Example
- periphery scan (no unused code)
- Tests/SizeReport/measure_size.sh

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
+++ b/Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
@@ -2,13 +2,6 @@
 import Foundation
 
 internal actor TxnSessionManager {
-    private enum Keys {
-        static let tagId = "ROKT_TXN_TAG_ID"
-        static let sessionId = "ROKT_TXN_SESSION_ID"
-        static let token = "ROKT_TXN_SESSION_TOKEN"
-        static let expiresAt = "ROKT_TXN_TOKEN_EXPIRES_AT"
-    }
-
     private let clock: () -> Date
 
     // nil disables persistence (in-memory only), preserving the lightweight test setup.
@@ -75,29 +68,24 @@ internal actor TxnSessionManager {
         token = nil
         expiresAt = nil
         guard let store else { return }
-        store.removeValue(forKey: Keys.tagId)
-        store.removeValue(forKey: Keys.sessionId)
-        store.removeValue(forKey: Keys.token)
-        store.removeValue(forKey: Keys.expiresAt)
+        TxnSessionPersistence.clear(store: store)
     }
 
     private var hasExpired: Bool {
-        guard let expiresAt else { return true }
-        return clock() >= expiresAt
+        TxnSessionPersistence.isExpired(expiresAt: expiresAt, clock: clock)
     }
 
     private func restoreFromStore() {
         guard let store, let roktTagId else { return }
         // Only restore a session bound to the current tag id; otherwise start clean.
-        guard store.string(forKey: Keys.tagId) == roktTagId else {
+        guard store.string(forKey: TxnSessionStoreKeys.tagId) == roktTagId else {
             clear()
             return
         }
-        sessionId = store.string(forKey: Keys.sessionId)
-        token = store.string(forKey: Keys.token)
-        expiresAt = store.string(forKey: Keys.expiresAt)
-            .flatMap(Double.init)
-            .map { Date(timeIntervalSince1970: $0/1000) }
+        let raw = TxnSessionPersistence.readRaw(store: store)
+        sessionId = raw.sessionId
+        token = raw.token
+        expiresAt = raw.expiresAt
         // Drop a persisted-but-expired token so we never start with stale state;
         // an expired JWT is dead server-side and a fresh session is minted at init.
         if hasExpired {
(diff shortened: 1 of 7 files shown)


Sources/Rokt_Widget/Rokt.swift
```diff
@@ -251,21 +251,45 @@ internal import RoktUXHelper
...
     /// e.g. WebView, and you want the session to be consistent across integrations.
     ///
     /// - Note: Empty strings are ignored and will not update the session.
+    /// - Note: Prefer ``setSession(_:)`` so the session token is also applied for offers and events.
     ///
     /// - Parameters:
     ///   - sessionId: The session id to be set. Must be a non-empty string.
+    @available(*, deprecated, message: "Use setSession(_:) to set session id and session token.")
     public static func setSessionId(sessionId: String) {
         shared.roktImplementation.setSessionId(sessionId: sessionId)
```
thomson-t, line 288: Should we add a warning log to indicate this is no longer supported?

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -1330,6 +1330,40 @@ class RoktInternalImplementation {
         }
     }
 
+    func setSession(_ session: RoktSession) {
+        guard let roktTagId,
+              !session.sessionId.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty,
+              !session.sessionToken.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        else {
+            return
```
thomson-t, line 1341: Log before return?

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -1330,6 +1330,40 @@ class RoktInternalImplementation {
...
+            roktTagId: roktTagId,
+            sessionId: session.sessionId,
+            sessionToken: sessionToken
+        )
+        sessionManager.updateSessionId(newSessionId: session.sessionId)
+    }
+
+    func getSession() -> RoktSession? {
+        guard let roktTagId else {
+            return nil
```
thomson-t, line 1365: Warning logs when not returning a session?

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -1330,6 +1330,40 @@ class RoktInternalImplementation {
...
 
+    func setSession(_ session: RoktSession) {
+        guard let roktTagId,
+              !session.sessionId.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty,
+              !session.sessionToken.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty
+        else {
+            return
+        }
+
+        let sessionToken = TxnSessionToken(token: session.sessionToken, expiresAt: session.expiresAt)
```
thomson-t, line 1344: Should we seed it if it is already expired?

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -1330,6 +1330,40 @@ class RoktInternalImplementation {
...
+        guard let snapshot = TxnSessionPersistence.load(roktTagId: roktTagId),
+              let sessionId = snapshot.sessionId,
+              !sessionId.isEmpty,
+              let token = snapshot.token,
+              !token.isEmpty,
+              let expiresAt = snapshot.expiresAt
+        else {
+            return nil
+        }
+        let expiresAtMs = Int64((expiresAt.timeIntervalSince1970 * 1000).rounded(.down))
```
thomson-t, line 1397: Should we still send if if it is expired?

Sources/Rokt_Widget/Rokt.swift
BrandonStalnaker: The deprecation will make any use show up as a warning in xCode but why not?

Sources/Rokt_Widget/RoktInternalImplementation.swift
BrandonStalnaker: good call out

Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
```diff
@@ -75,29 +68,24 @@ internal actor TxnSessionManager {
...
-        guard let expiresAt else { return true }
-        return clock() >= expiresAt
+        TxnSessionPersistence.isExpired(expiresAt: expiresAt, clock: clock)
     }
 
     private func restoreFromStore() {
         guard let store, let roktTagId else { return }
         // Only restore a session bound to the current tag id; otherwise start clean.
-        guard store.string(forKey: Keys.tagId) == roktTagId else {
+        guard store.string(forKey: TxnSessionStoreKeys.tagId) == roktTagId else {
```
denischilik, line 81: The code looks good to me. Just one small nit: this tag check could be moved into TxnSessionPersistence as a dedicated helper (e.g. isTagMatch or hasMatchingTag). The same logic already exists in load(), so extracting it would avoid duplication and keep all direct access to TxnSessionStoreKeys in one place.

denischilik (approved):

thomson-t (approved):

PR #276, opened 2026-08-03
feat(diagnostics): log public API usage as INFO diagnostics

thomson-t:
## Background

We currently have no visibility into which public SDK APIs partners actually call, how often, or how they integrate (SwiftUI `RoktLayout` vs the imperative `selectPlacements`). This adds a lightweight INFO diagnostic on every partner-facing public API call so usage is observable, plus non-PII hints about the parameters passed. It reuses the existing `sendDiagnostics` pipeline — no new networking and no backend change.

Fixes N/A

## What Has Changed

- New `RoktAPIHelper.logApiCalled(_:_:)` helper that emits an INFO diagnostic with a distinct code per public API. Fire-and-forget; self-guards on `roktTagId` so it never throws into or blocks the partner call.
- Instrumented the public API surface with distinct codes: `init`, `selectPlacements`, the SwiftUI `RoktLayout`, `selectShoppableAds`, `purchaseFinalized`, `events`, `globalEvents`, `close`, `registerPaymentExtension`, `setCustomBaseURL`, `setFrameworkType`, `getSessionId`, `setSessionId`, `handleURLCallback`, and `setBuiltInPayPalRedirectURLScheme`.
- `RoktLayout` (SwiftUI) and `selectPlacements` get separate codes, so the declarative-vs-imperative integration split is visible.
- Calls that normally run before `init` (`setCustomBaseURL`, `setFrameworkType`, `globalEvents`) are buffered and flushed once `init` sets the tag id, so they still get attributed. The buffer is bounded and thread-safe.
- `additionalInformation` carries non-PII hints only (booleans/counts/enums as strings): e.g. `hasConfig`, `colorMode`, `cacheEnabled`, `embedded`, `attributeCount`, `frameworkType`, `success`. Never logged: attribute values, view names, URLs, session ids, the redirect URL, or payment-extension config/secrets.

## How Has This Been Tested

- New unit tests in `TestPublicApiDiagnostics.swift` (4 tests): a representative public call emits the expected code at INFO severity; the color-mode mapping; and the pre-init buffer routing (buffered before init, sent inline after). All pass.
- `xcodebuild test -scheme Rokt-Widget` on iOS Simulator: build succeeds, tests pass, no new warnings from these changes.

## Notes

Logging reuses the existing diagnostics endpoint and severity levels; there is no change to the request contract or any backend service.

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/RoktInternalImplementation.swift
+++ b/Sources/Rokt_Widget/RoktInternalImplementation.swift
@@ -304,6 +304,17 @@ class RoktInternalImplementation {
         }
     }
 
+    /// Bounded, non-PII format guard for the mParticle public-API diagnostic codes forwarded by
+    /// the Rokt kit. Accepts uppercase SNAKE_CASE identifiers only (e.g. `LOG_EVENT`), 1...40
+    /// chars — which structurally excludes event/screen names, attribute values, URLs, and ids
+    /// from ever reaching the `code` tag. Malformed codes are dropped by the caller.
+    static func isValidMParticleApiCode(_ code: String) -> Bool {
+        guard (1...40).contains(code.count), let first = code.first, ("A"..."Z").contains(first) else {
+            return false
+        }
+        return code.allSatisfy { ("A"..."Z").contains($0) || ("0"..."9").contains($0) || $0 == "_" }
+    }
+
     /// Log a public API call, buffering it until init when the SDK isn't initialised yet. Only
     /// `setCustomBaseURL` / `setFrameworkType` need this (they run pre-init); everything else logs inline.
     func logApiCallBuffered(_ code: String, _ additionalInfo: [String: String] = [:]) {
--- a/Sources/Rokt_Widget/RoktInternalImplementation.swift
+++ b/Sources/Rokt_Widget/RoktInternalImplementation.swift
@@ -304,8 +304,8 @@ class RoktInternalImplementation {
         return code.allSatisfy { ("A"..."Z").contains($0) || ("0"..."9").contains($0) || $0 == "_" }
     }
 
-    /// Log a public API call, buffering it until init when the SDK isn't initialised yet. Only
-    /// `setCustomBaseURL` / `setFrameworkType` need this (they run pre-init); everything else logs inline.
+    /// Log a public API call, buffering it until init when the SDK isn't initialised yet. This supports
+    /// `globalEvents`, mParticle API forwarding, and configuration APIs that may run before init.
     func logApiCallBuffered(_ code: String, _ additionalInfo: [String: String] = [:]) {
         let shouldSend = pendingApiLogsLock.withLock {
             guard roktTagId == nil else { return true }
--- a/Sources/Rokt_Widget/RoktInternalImplementation.swift
+++ b/Sources/Rokt_Widget/RoktInternalImplementation.swift
@@ -11,6 +11,24 @@ class RoktInternalImplementation {
     private static let notInitializedDiagnosticCode = "[NOT_INITIALIZED]"
     private static let cacheHitDiagnosticCode = "[CACHE_HIT]"
     private static let cacheHitMessage = "Cache hit for view - %@"
+    // Public-API-usage diagnostics (INFO severity). Keep this a small, bounded set of codes.
+    static let apiInitCode = "[API_INIT]"
+    static let apiInitMParticleCode = "[API_INIT_MPARTICLE]"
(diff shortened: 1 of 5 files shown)


Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -11,6 +11,24 @@ class RoktInternalImplementation {
     private static let notInitializedDiagnosticCode = "[NOT_INITIALIZED]"
     private static let cacheHitDiagnosticCode = "[CACHE_HIT]"
     private static let cacheHitMessage = "Cache hit for view - %@"
+    // Public-API-usage diagnostics (INFO severity). Keep this a small, bounded set of codes.
+    static let apiInitCode = "[API_INIT]"
```
denischilik, line 15: Nit: consider extracting these into a dedicated enum (e.g. `PublicApiDiagnosticCode`) instead of a pile of `static let`s on `RoktInternalImplementation`. Then type `logApiCalled` / `logApiCallBuffered` against that enum rather than `String`, so call sites can’t pass an arbitrary code by mistake.

```swift
enum PublicApiDiagnosticCode: String {
    case apiInit = "[API_INIT]"
    case apiInitMParticle = "[API_INIT_MPARTICLE]"
    // ...
}
```

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -260,6 +288,43 @@ class RoktInternalImplementation {
 
     func setFrameworkType(_ frameworkType: RoktFrameworkType) {
         self.frameworkType = frameworkType
+        logApiCallBuffered(Self.apiSetFrameworkTypeCode, ["frameworkType": frameworkType.toString])
+    }
+
+    // MARK: - Public API usage diagnostics
+
+    /// Map a partner-supplied RoktConfig color mode to a bounded, non-PII string tag.
+    static func colorModeString(_ config: RoktConfig?) -> String {
```
denischilik: Nit: this mapping belongs with `RoktConfig` / `ColorMode`, not on `RoktInternalImplementation`. Prefer an extension, e.g.:

```swift
extension RoktConfig {
    /// Bounded, non-PII tag for diagnostics.
    var colorModeDiagnosticValue: String { ... }
}
```

(or the same idea on `ColorMode` itself).

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -260,6 +288,43 @@ class RoktInternalImplementation {
...
+        case .dark: return "dark"
+        case .system: return "system"
+        @unknown default: return "unknown"
+        }
+    }
+
+    /// Log a public API call, buffering it until init when the SDK isn't initialised yet. Only
+    /// `setCustomBaseURL` / `setFrameworkType` need this (they run pre-init); everything else logs inline.
+    func logApiCallBuffered(_ code: String, _ additionalInfo: [String: String] = [:]) {
+        guard roktTagId == nil else {
```
denischilik: Bug: `roktTagId == nil` is checked outside the lock, then the entry is appended under the lock. If `init` sets `roktTagId` and calls `flushPendingApiLogs()` between that check and the append, the log lands in the buffer after the flush and is never sent.

Please re-check `roktTagId` under the same lock (or otherwise serialize the tag-id / flush / buffer path) so a concurrent pre-init call can’t orphan a log.

denischilik (commented):
A few nits plus one concurrency bug in the pre-init buffer path.

Sources/Rokt_Widget/RoktInternalImplementation.swift
thomson-t: This would need to refactor the existing codes as well. We can do it as part of another PR.

Sources/Rokt_Widget/RoktInternalImplementation.swift
thomson-t: Updated

Sources/Rokt_Widget/RoktInternalImplementation.swift
thomson-t: Fixed

Sources/Rokt_Widget/RoktInternalImplementation.swift
denischilik: Sure

denischilik (approved):

BrandonStalnaker (dismissed):
A few questions but all nits and looks good in general

BrandonStalnaker (approved):
LGTM

PR #278, opened 2026-08-05
fix: use Pact branch selector in Release Draft gate

denischilik:
## Background
Release – Draft fails with `No version with tag main exists` because `can-i-deploy --latest main` looks up a Pact tag, while publish only sets branch `main`.

## What Has Changed
- Switched the Release Draft contract gate to `--latest --branch main` so it matches how consumer pacts are published

## Checklist
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have tested this locally.

### DIFF
--- a/.github/workflows/release-draft.yml
+++ b/.github/workflows/release-draft.yml
@@ -23,8 +23,8 @@ jobs:
   contract-verified:
     name: Contract verified vs production
     runs-on: ubuntu-latest
-    # Blocking: refuse to open a release PR unless the latest main consumer
-    # contract is verified against the provider currently in production.
+    # Blocking: refuse to open a release PR unless the latest consumer
+    # contract for this branch is verified against production.
     environment: pact-publish
     steps:
       - name: Check provider verification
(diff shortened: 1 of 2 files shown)


.github/workflows/release-draft.yml
```diff
@@ -43,7 +43,7 @@ jobs:
             "${PACT_CLI_IMAGE}" \
             broker can-i-deploy \
               --pacticipant rokt-sdk-ios \
-              --latest main \
+              --latest --branch main \
```
thomson-t: This can be triggered from a maintenance branch as well. 
Can we change it like, add an environment 
`PACT_BRANCH: ${{ github.ref_name }}`
then
`--branch "${PACT_BRANCH}" \` ?

We would also need to update `pact-consumer.yml` and include 
```
branches:
      - main
      - maintenance/*
```
then remove the condition for publishing to pact broker `github.ref == 'refs/heads/main'`

denischilik: Good call — updated both sides:

- `release-draft.yml`: `PACT_BRANCH: ${{ github.ref_name }}` + `--latest --branch "${PACT_BRANCH}"`
- `pact-consumer.yml`: publish on `main` and `maintenance/*`, dropped the `refs/heads/main` publish condition

Same changes on the Android twin: https://github.com/ROKT/sdk-android-source/pull/1135

thomson-t (approved):

PR #284, opened 2026-08-14
fix: multicast Rokt.globalEvents so kit and client can both subscribe

jamesnrokt:
## Background

`Rokt.globalEvents(onEvent:)` stored its subscriber in a single dictionary slot, so the **last** caller to subscribe silently unsubscribed everyone before it:

```swift
// RoktInternalImplementation.swift (before)
if isGlobal, viewName.isEmpty {
    roktEventMap[Self.defaultRoktInitEvent] = onEvent   // overwrite, not append
}
```

There are always at least two subscribers in an mParticle integration, and they collide:

1. The **mParticle Rokt kit** subscribes during `didFinishLaunchingWithConfiguration:` and uses `InitComplete(success: true)` as the sole trigger to mark itself started ([`MPKitRokt.m:80-93`](https://github.com/mParticle/mparticle-apple-sdk/blob/main/kits/rokt/rokt/Sources/mParticle-Rokt/MPKitRokt.m)).
2. The **host app** subscribes either directly via `Rokt.globalEvents`, or through `MParticle.rokt.globalEvents:`, which the kit forwards straight back into the same slot ([`MPKitRokt.m:555-558`](https://github.com/mParticle/mparticle-apple-sdk/blob/main/kits/rokt/rokt/Sources/mParticle-Rokt/MPKitRokt.m)).

Whichever registers second wins. The failure is severe and completely silent when the app wins:

- the kit never receives `InitComplete`, so `started` stays `NO`;
- `MPKitContainer.isActiveAndNotDisabled:` therefore excludes Rokt from `activeKitsRegistry`;
- `forwardSDKCall:` iterates an empty list, so **every `selectPlacements` call is discarded without ever invoking the caller's `onEvent`**.

The partner sees no placements *and* no events at all — not even `PlacementFailure`, because the call never reaches the Rokt SDK, which would otherwise always report one via `preExecuteFailureHandler()`. There is no log on the app side and no recovery for the rest of the process lifetime.

This was found while investigating a partner reporting missing placements with zero Rokt events of any type.

Fixes [(issue)]

## What Has Changed

Global event subscribers are now multicast instead of single-slot.

- Replaced the shared `roktEventMap[DEFAULT_ROKT_INIT_EVENT]` slot with a dedicated `globalEventListeners` array guarded by an `NSLock` — `mapEvents` is reachable from any thread while delivery happens on main.
- `mapEvents(isGlobal: true, onEvent:)` now **appends** a subscriber; passing `nil` clears all of them, preserving the previous "stop delivering" behaviour.
- Both `InitComplete` emit sites (`handleInitSuccess` / `handleInitFailure`) fan out to every subscriber.
- Removed the now-unused `defaultRoktInitEvent` sentinel key.

Per-view listeners registered through `Rokt.events(identifier:onEvent:)` are deliberately untouched — they keep today's replace-on-re-register semantics, and a test pins that.

**No public API change.** `Rokt.globalEvents(onEvent:)` and `Rokt.events(identifier:onEvent:)` keep their exact signatures and ObjC selectors; only the internal dispatch changes. Existing single-subscriber integrations behave identically.

### Behaviour change to be aware of

A caller that subscribes to `globalEvents` more than once now receives each event once **per subscription**, where previously the earlier subscription was silently dropped. `globalEvents` is documented as a process-wide subscription and the kit registers under a `dispatch_once`, so this is not expected to affect existing integrations — but it is the one observable difference and worth a reviewer's eye.

## How Has This Been Tested

New `Tests/Rokt_WidgetTests/TestGlobalEventsMulticast.swift`, driving init deterministically through the `makeTxnInitServiceOverride` seam with a stubbed HTTP client (same pattern as `TestTxnInitWiring`), against a fresh `RoktInternalImplementation` rather than the shared singleton:

- `test_globalEvents_bothKitAndClientReceiveInitComplete` — two subscribers, both receive exactly one `InitComplete(success: true)`. Fails on `main` (the first subscriber gets nothing).
- `test_globalEvents_bothSubscribersReceiveInitFailure` — same fan-out on the failure path.
- `test_globalEvents_nilOnEventRemovesAllSubscribers` — `nil` clears every subscriber.
- `test_globalEvents_perViewListenersAreUnaffected` — a per-view listener does not receive global `InitComplete`.

`TestRokt.setUp` now clears global subscribers between tests. Those two existing tests subscribe on the shared `Rokt.shared` singleton, so without a reset the multicast list would leak across tests and fulfil expectations belonging to already-finished cases.

Validation run:

- `trunk check` — clean on all changed files.
- `xcodebuild test -scheme Rokt-Widget -only-testing:Rokt_WidgetTests` — full unit suite green on iPhone 16 Pro (iOS 18.6).

Not run: `periphery scan` and `Tests/SizeReport/measure_size.sh`. The change is a few lines in one file with no new types or assets, so size impact is nil, but flagging that I skipped those two steps rather than implying they passed.

## Notes

This fixes the *collision*. It does not fix the underlying fragility that the mParticle Rokt kit only becomes active on `InitComplete(success: true)` — a failed Rokt init (network error, 429, font failure) still leaves the kit permanently inactive for the process lifetime, with the same silent-drop symptom. That is addressed separately by adding init retry, stacked on this branch.

Related follow-up worth considering, not included here: per-view listeners registered via `Rokt.events(identifier:)` have the same single-slot overwrite behaviour, scoped per `viewName`. Lower impact, but the same class of bug.

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/RoktInternalImplementation.swift
+++ b/Sources/Rokt_Widget/RoktInternalImplementation.swift
@@ -31,7 +31,6 @@ class RoktInternalImplementation {
     private static let maxPendingApiLogs = 10
     private static let initFailedError = "INIT_FAILED"
     private static let fontFailedError = "FONT_FAILED"
-    private static let defaultRoktInitEvent = "DEFAULT_ROKT_INIT_EVENT"
     private static let trackingConsentError = "tracking consent not authorised"
     private static let cacheDurationKey = "cacheDuration"
     private static let cacheAttributesKey = "cacheAttributeKeys"
@@ -101,6 +100,11 @@ class RoktInternalImplementation {
     // store callback for partner event integration
     private var roktEvent: ((RoktEvent) -> Void)?
     private var roktEventMap: [String: ((RoktEvent) -> Void)?] = [:]
+
+    // Multicast: the mParticle kit and the host app both subscribe through Rokt.globalEvents,
+    // so a single slot let whichever registered last silently unsubscribe the other.
+    private var globalEventListeners: [(RoktEvent) -> Void] = []
+    private let globalEventListenersLock = NSLock()
 
     // debounce work item for EmbeddedSizeChanged
     private var sizeChangeWorkItem: DispatchWorkItem?
@@ -899,6 +903,25 @@ class RoktInternalImplementation {
         }
     }
 
+    private func addGlobalEventListener(_ onEvent: @escaping (RoktEvent) -> Void) {
+        globalEventListenersLock.lock()
+        defer { globalEventListenersLock.unlock() }
+        globalEventListeners.append(onEvent)
+    }
+
+    func removeAllGlobalEventListeners() {
+        globalEventListenersLock.lock()
+        defer { globalEventListenersLock.unlock() }
+        globalEventListeners.removeAll()
+    }
+
+    private func sendEventToGlobalListeners(_ roktEvent: RoktEvent) {
+        globalEventListenersLock.lock()
+        let listeners = globalEventListeners
+        globalEventListenersLock.unlock()
+        listeners.forEach { $0(roktEvent) }
+    }
+
     func initWith(
         roktTagId: String,
         mParticleKitDetails: MParticleKitDetails?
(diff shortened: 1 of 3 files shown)


thomson-t (dismissed):

BrandonStalnaker (dismissed):
I could see adding a test that confirms “N subscriptions → N callbacks.” or noting that accumulation in some way but it's just a nit and could be added later. Looks good otherwise and good catch.

BrandonStalnaker (approved):

PR #285, opened 2026-08-14
fix: recover from a failed init instead of staying uninitialised

jamesnrokt:
## Background

A failed init leaves the SDK dead for the rest of the process. `TxnInitService` already retries 5xx and transport blips *within* a single request, but once that request gives up, nothing tries again — `isInitialized` stays `false` and every later `selectPlacements` call is discarded until the app is restarted.

The two realistic ways to hit this are a cold start with no network (init fires before connectivity is up) and a 429 during a traffic burst — both transient, both currently terminal for the session.

## What Has Changed

- Init failures worth retrying now schedule up to **three bounded recovery attempts at 2s, 8s and 30s**. Recoverable means `429`, any `5xx`, or an `NSURLErrorDomain` transport failure.
- Non-recoverable failures — a wrong tag id, a rejected request, any other 4xx — are **not** retried; they fail the same way every time.
- The attempt counter resets on a successful init and on every fresh `initWith` call, so a later init isn't penalised by an earlier failure.
- Recovery is tied to the existing `initGeneration`, so a superseded init can't resurrect itself, and a scheduled attempt is skipped if init has since succeeded.
- Retry scheduling goes through an injectable `initRecoveryScheduler` (defaults to `DispatchQueue.main.asyncAfter`) so tests don't wait on real delays.
- Diagnostics behaviour is unchanged: 429s still aren't reported; the `InitComplete(success: false)` event still fires on each failure, including before a recovery attempt.

No public API change.

## How Has This Been Tested

New `TestInitRecovery` suite (5 tests) driving a stubbed HTTP client:

- retries after a 429 and eventually succeeds — emits `InitComplete(false)` then `InitComplete(true)`, 2 requests
- retries after a 503
- does not retry a 400
- stops after exhausting attempts — 4 requests total (initial + 3 recoveries)
- counter resets on a fresh `initWith`

Full suite locally: **703 tests, 0 failures** (2 skipped), `xcodebuild test -scheme Rokt-Widget` on iPhone 17 Pro sim. `trunk check` clean.

## Notes

Client-visible behaviour on a transient failure: the app receives `InitComplete(success: false)` up to four times before the SDK recovers, rather than once. Existing listeners that treat `InitComplete(false)` as terminal will still work, but they'll now see the SDK become usable afterwards.

The retry curve is fixed, not jittered — if a burst rate-limits many devices at once they all come back at 2s/8s/30s together. Worth a reviewer eye on whether 3 attempts at that spacing is the right ceiling for a 429 burst, since every host retrying on the same curve is itself a load pattern.

Related to the mParticle init-storm work. Sibling branch `fix/global-events-multicast` touches `handleInitFailure` too; no overlapping lines after this rebase, but whichever lands second needs a trivial merge there.

## Checklist

### DIFF
--- a/Sources/Rokt_Widget/RoktInternalImplementation.swift
+++ b/Sources/Rokt_Widget/RoktInternalImplementation.swift
@@ -1034,12 +1034,26 @@ class RoktInternalImplementation {
     }
 
     // A wrong tag id or a rejected request will fail the same way every time, so only
-    // rate limiting, server faults and transport failures are worth coming back for.
+    // rate limiting, server faults and transient transport failures are worth coming back for.
+    // The URL error allowlist matches TxnEventService / ForwardPaymentRetryRules: a cancelled
+    // request, a bad URL or a TLS/ATS rejection is permanent and must not be retried.
     private static func isRecoverable(error: Error, statusCode: Int?) -> Bool {
         if let statusCode {
             return statusCode == rateLimitedStatusCode || (500..<600).contains(statusCode)
         }
-        return (error as NSError).domain == NSURLErrorDomain
+        let nsError = error as NSError
+        guard nsError.domain == NSURLErrorDomain else { return false }
+        switch nsError.code {
+        case NSURLErrorTimedOut,
+             NSURLErrorNetworkConnectionLost,
+             NSURLErrorNotConnectedToInternet,
+             NSURLErrorCannotConnectToHost,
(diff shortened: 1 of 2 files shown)


BrandonStalnaker (commented):
Generally looks good to me but I have minor suggestion I'd appreciate you taking a look at.

BrandonStalnaker (approved):

thomson-t (approved):

PR #298, opened 2026-08-19
feat: add clearSession() for self-service terminals

nickolas-dimitrakas:
## Problem

On a shared self-service terminal — a kiosk, counter tablet, or shared POS device — a queue of unrelated customers uses one device, and every one of them was landing in the same Rokt session.

Session identity is carried solely by a JWT the SDK persists in `UserDefaults` and replays as `Authorization: Bearer …` on every offers and events request. The gateway continues the existing session whenever a live token is presented, and mints a new one only when it is absent. Because every offers/events response rotates the token and pushes its expiry forward, a busy terminal never lets it lapse — so the session grew indefinitely, one page per customer.

The practical effect: frequency capping, dedupe, attribution and reporting all treated consecutive strangers as one person. Any two customers arriving within one token lifetime of each other were merged.

`setSessionId`/`getSessionId` do not help here — they drive a separate legacy diagnostics session id and never touch the JWT.

## Change

Adds `Rokt.clearSession()`, a public API that ends the current session so the next `selectPlacements` starts a new one.

Ordering is the load-bearing part. `EventQueue.flush()` runs first and synchronously, and `dispatchTxnEvents` builds its `TxnEventService` (and that service's `TxnSessionManager`) before spawning the send task — so the departing customer's buffered events still go out under the departing customer's token. Only then is session state wiped, synchronously, so a placement issued immediately afterwards cannot rehydrate what was just dropped.

`clearSession()` touches session-scoped state only:

| Cleared | Why |
|---|---|
| Persisted txn session + token | The session identity itself |
| Legacy diagnostics session id | Also a session id |
| Real-time event store | Session-scoped; already cleared on any session-id change today |
| Experience cache | Fetched inside the session being dropped, and a cache hit skips the offers call that mints the new one |

### Supporting changes

- **Session epoch.** A counter in the txn store, captured by each `TxnSessionManager` at init and re-checked before it writes or clears. An offers/events response that lands after a reset can no longer restore the session that was just dropped. Housekeeping clears (tag-id mismatch, expired-on-restore) deliberately do not bump it, so a healthy in-flight request is not fenced out.
- **Session-bound event replay.** `TxnPendingEventBatch` now records the session it belongs to, and replay sends unauthenticated with an explicit `session_id`, so a previous customer's events stay on the previous customer's session instead of attaching to whoever is at the terminal now. Batches written by an older SDK version carry no binding and are dropped rather than misattributed.
- **One-shot cache-read bypass.** Clearing the cache is asynchronous while the cache read is a direct synchronous file read, so clearing alone would race the very next placement. The flag makes that first placement deterministic. Cache writes are unaffected, so the new customer's experience is still cached for their own session.

## Backwards compatibility

Additive. Nothing changes for existing integrations unless `clearSession()` is called. `TxnPendingEventBatch.sessionId` is optional, so files written by the previous version still decode, and a downgrade can still read files written by this one.

`Rokt.getSessionId()` returns `nil` after a reset — documented on the API, and relevant to WebView session hand-off.

## Testing

747 tests, 0 failures (`Rokt-Widget` scheme, iOS 26.3 simulator). `trunk check` clean across 207 files.

New coverage includes: the persisted txn session is actually dropped; a response landing after a reset does not resurrect the session; a stale 401 does not tear down the live session; an in-flight sender keeps the departing token; pending batches replay against their own session and unbound ones are dropped; and `clearSession()` with no active session is a no-op.

## Follow-ups (not in this PR)

- mParticle kit and mParticle core changes, which call this automatically on MPID change — held pending review of this PR.
- Android equivalents, to be mirrored once this logic is settled.
- The gateway's handling of an unauthenticated events batch carrying an explicit `session_id` is relied upon by the replay path and is worth confirming with the transactions team.

## Manual E2E verification (simulators, production gateway)

Ran a before/after matrix on two iOS simulators (iPhone 17 Pro = "kiosk", iPhone Air = "regular"), against the production gateway with a bottomsheet test layout on an internal test account. Each transaction = a full `selectPlacements` → offers → render → events cycle; session identity was read from the SDK's persisted store after every transaction, and all traffic was inspected via a local proxy. The kiosk boundary was driven through a temporary button wired to `Rokt.clearSession()` — the same single funnel the mParticle kit triggers call.

| Run | Transactions | Result |
|---|---|---|
| Kiosk, **base** (main) | customer A → customer B, no reset available | Both landed in **one session** — the defect this PR fixes, reproduced |
| Regular, **base** (main) | two placements, same user | One session (expected) |
| Kiosk, **this PR** | customer A → `clearSession()` → customer B | **Two distinct sessions** — fix confirmed |
| Regular, **this PR** | two placements, no reset | **One session — nothing severed prematurely** |
| Regular, **this PR** | third placement after full app kill + relaunch | **Same session — persistence across restart intact** |

Additional observations:

- After `clearSession()`, all four `ROKT_TXN_*` keys and the legacy diagnostics session id were removed from `UserDefaults`, and the epoch counter was bumped — verified on disk before the next placement ran.
- On the regular-flow simulator the epoch key was never written at all: the reset machinery is provably inert unless `clearSession()` is called.
- The next placement after a reset went out with no `Authorization` header and the gateway minted a fresh session, matching the design.


Sources/Rokt_Widget/DataAccess/TxnSessionManager.swift
```diff
@@ -57,6 +66,7 @@ internal actor TxnSessionManager {
     }
 
     func update(sessionId: String, sessionToken: TxnSessionToken) {
+        guard isCurrentEpoch else { return }
         self.sessionId = sessionId
         token = sessionToken.token
         expiresAt = sessionToken.expiresAtDate
```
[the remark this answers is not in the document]
nickolas-dimitrakas: Good catch — confirmed and fixed in 485e63f.

You're right that actor isolation doesn't cover this. `clearPersistedSession` is deliberately `nonisolated` (it has to take effect synchronously before the caller's next `selectPlacements`, so it can't hop onto the actor), which means it doesn't queue behind actor-isolated work and could interleave between the epoch check and `persist` inside `update` — writing the departing session back alongside the new epoch, and silently undoing the reset in precisely the in-flight-response case the fence exists for.

Fix: a single static `NSLock` now guards every read-modify-write of the shared store, taken exactly once per public entry point (`update` x2, `clear`, `clearPersistedSession`, `restoreFromStore`). Internal helpers are renamed `*Locked` to make the "caller already holds the lock" contract explicit, since `NSLock` isn't reentrant. The lock is static so it serialises instances against each other and against the nonisolated entry point.

Worth noting the Android SDK doesn't have this class of bug — there the reset goes through the same `synchronized(lock)` instance as `updateToken`, with no nonisolated bypass.

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -1336,6 +1380,19 @@ class RoktInternalImplementation {
...
+    /// `clearSession()` empties the cache asynchronously, but this read is a direct synchronous
+    /// file read — so the one-shot flag is what makes the first placement after a reset
+    /// deterministically reach the network.
+    private func shouldReadFromCache() -> Bool {
+        if mustBypassCacheOnNextExecute {
+            mustBypassCacheOnNextExecute = false
+            return false
+        }
+        return isCacheEnabledAndConfigured()
+    }
```
BrandonStalnaker: View-state cache can still attach the previous customer to the next placement

This kinda confusing to explain but clearSession() correctly refuses a experience cache hit via mustBypassCacheOnNextExecute, because ExperienceCacheManager.clearCache() deletes files asynchronously while getCachedExperienceResponse is a sync file read.

Here the flag is consumed at the start of execute.

Sources/Rokt_Widget/RoktInternalImplementation.swift
BrandonStalnaker: View-state cache can still attach the previous customer to the next placement Part 2 of 2

Later you can see here that after the new offers response, processLayoutPageExecutePayload still reads the view-state cache with isCacheEnabledAndConfigured() — not the bypass.

If those files are still on disk (very likely: delete is async; the flag is already false), the next customer can inherit sentEventHashes (impressions/events suppressed as “already sent”) and plugin view states (previous customer’s UI progress).
I think that undercuts the session reset for any partner with cache enabled.

BrandonStalnaker (changes requested):

Sources/Rokt_Widget/RoktInternalImplementation.swift
```diff
@@ -100,7 +101,15 @@ class RoktInternalImplementation {
...
-    private let txnPendingEventStore: TxnPendingEventStoring = TxnPendingEventStore()
+    // Test-only override; production always uses the default file-backed store.
+    var txnPendingEventStore: TxnPendingEventStoring = TxnPendingEventStore()
+
+    // Backing store for the txn session. Test-only override; production uses UserDefaults.
+    var txnSessionStore: TxnSessionStore = UserDefaultsTxnSessionStore()
+
+    // Set by clearSession, consumed by the next cache-read decision. A cache hit satisfies a
+    // placement without a network call, and the server is what mints a session.
+    private var mustBypassCacheOnNextExecute = false
```
thomson-t: Is this required? We are clearing the cache files from `clearSession`, isn't it?

nickolas-dimitrakas: Yes, still needed — we clear the cache, but we can't guarantee the delete lands before the next offers call.

`ExperienceCacheManager.clearCache()` deletes asynchronously (barrier on a concurrent queue), while the cache read is a direct synchronous `Data(contentsOf:)`. So a `selectPlacements` immediately after `clearSession()` — which is exactly the kiosk pattern — can still find the file.

Clearing alone would work most of the time, which is the worst outcome to debug. The flag makes that first placement deterministic.

Sources/Rokt_Widget/RoktInternalImplementation.swift
nickolas-dimitrakas: You're right, and this was a real gap — fixed in d7f28b9.

The flag was consumed at the experience-cache read, so by the time `processLayoutPageExecutePayload` reached the view-state read it was already false and that read went ahead. Exactly as you say: the delete is async, the files are usually still there, and the next customer inherits `sentEventHashes` (impressions suppressed as already sent) and plugin view states.

Fix is to latch the decision once per execute rather than consume it at the first read, so both reads see the same answer. Both now go through `shouldReadFromCache()`. The cache *write* is deliberately left on `isCacheEnabledAndConfigured()` so the new customer's experience still caches for their own session.

Good catch — this is the half of the cache problem I'd left behind.

Sources/Rokt_Widget/RoktInternalImplementation.swift
nickolas-dimitrakas: Answered on part 2 — fixed in d7f28b9. The flag is now latched once per execute instead of consumed at the first read, so the experience-cache read and the view-state read both honour it.

BrandonStalnaker (approved):

thomson-t (approved):

PR #302, opened 2026-08-21
ci: enforce declared dependencies at compile time and in CI

nickolas-dimitrakas:
## Background

Follow-up to #300 / #301. CI never caught the undeclared `DcuiSchema` dependency, and this closes that gap with two layers: the compiler rejects the source-level mistake, and a script catches the manifest-level one.

The suggestion to add a `SPM_GENERATE_FRAMEWORK=1` build step does not work — that build passes with *and* without the dependency declaration, so it would be a green check proving nothing. It makes only the top-level product dynamic and links the transitive Swift modules statically into it. Reproducing the real failure needs every product built as its own dynamic framework, which requires patching third-party manifests, so it is not something to pin CI to.

## What Has Changed

**1. `MemberImportVisibility` on the `Rokt_Widget` target** (thomson-t's suggestion). Using a member of a module not imported in that file is now a compile error, so a type reaching us implicitly through a dependency's public API cannot be used by accident. Cost: 3 import lines across 109 files — `import Combine` in `RoktLayoutViewModel.swift` (pre-existing `ObservableObject`/`@Published` with no Combine import) and `internal import DcuiSchema` in the two `PaymentProvider` files, matching the existing `internal import RoktUXHelper` style so DcuiSchema stays out of our public interface.

**2. `scripts/check_declared_dependencies.sh`** — builds the target and fails on any package-graph module referenced by its undefined symbols but not declared for `Rokt_Widget` in `Package.swift`.

**3. `.github/workflows/pull-request.yml`** — `dependency-declaration-check` job, `needs: trunk-check`, same runner/Xcode as the other build jobs (~2m35s).

**4. `AGENTS.md`** — the import rule under "Always Do", and the script as step 6 of local validation.

CI YAML is listed in AGENTS.md as needing an explicit request; this was explicitly requested.

## Why both layers

They close different holes, and I verified the gap between them rather than assuming:

- **MIV does not enforce the manifest declaration.** With the imports in place I removed `dcui-swift-schema` from `Package.swift` and the build still **succeeded**. Under `xcodebuild` every package module's `.swiftmodule` lands in a shared `Build/Products` directory that sits on the import search path, so `import DcuiSchema` resolves whether or not it is declared.
- **The script does not see source intent**, only what actually gets linked.

So MIV catches "used a module you did not import"; the script catches "linked a module you did not declare".

## How Has This Been Tested

**MIV rejects the original bug.** Removing the new import reproduces it as a compile error:
```
Extensions/RoktUxEvent+Extension.swift:56:56: error: property 'rawValue' is not available
  due to missing import of defining module 'DcuiSchema'
```

**MIV needs nothing beyond those 3 imports.** Full build succeeds. The 13-file `RoktContracts` cluster (via `@_exported import` in `Rokt.swift`) and ~14 files using `DispatchQueue` without importing `Dispatch` all pass, since SE-0444 honours re-exports.

**Script red/green.** Green on the correct manifest (exit 0). Red with the declaration removed (exit 1, naming `DcuiSchema`). Note the build itself still *succeeds* in the failing case — which is the point: this catches what no build in CI surfaces.

**Regression test for the parsing fix.** The original `sed` anchored on the leading identifier and dropped 1004 of 1433 symbols, leaving detection resting on a single `.rawValue` call inside a log string. Rebuilt with both `.rawValue` sites removed, so only metadata and conformance symbols remain, and compared parsers on identical object files:
```
DcuiSchema symbols present:
  type metadata for DcuiSchema.PaymentProvider
  protocol conformance descriptor for DcuiSchema.PaymentProvider : Swift.RawRepresentable

OLD parser:  RoktContracts RoktUXHelper             <- false PASS
NEW parser:  DcuiSchema RoktContracts RoktUXHelper  <- exit 1, correct
```

**Also:** `trunk check` clean (shellcheck, shfmt, actionlint, yamllint, markdownlint, swiftformat, swiftlint). Script targets bash 3.2 — macOS runners have no bash 4. Two guards abort with an error rather than passing silently if the build layout changes and either the module set or the symbol set comes back empty; a check that cannot fail is worse than no check.

## Notes

- **Product vs module names:** `dump-package` yields product names, `nm` yields module names. All three dependencies are 1:1 today, and a future divergence surfaces as a loud false failure rather than a missed violation, so it is documented in a comment rather than hardened now.
- **Scope:** the `Rokt_Widget` target. `Packages/rokt-payment-extension-ios` and the 96 test files are untouched — enabling MIV there is a separate, larger job.
- MIV is not enforced on the CocoaPods path (the podspec sets no Swift flags and does not read `Package.swift`); CI builds via SPM, and the sources are identical either way.

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -89,6 +89,7 @@ You are a senior iOS SDK engineer specializing in stable, lightweight client lib
   3. Run unit tests: Use the `rokt-Example` scheme → Command + U (or `xcodebuild test ...`).
   4. Run periphery scan: `periphery scan` — ensure no unused code is introduced.
   5. If change affects code, assets, or dependencies: Run size report → `Tests/SizeReport/measure_size.sh` and confirm no unacceptable increase.
+  6. If the change touches imports, dependencies, or `Package.swift`: run `./scripts/check_declared_dependencies.sh` — fails when `Rokt_Widget` uses a module that reaches it only through a dependency's public API and is not declared in `Package.swift`.
   - Only propose / commit changes if all steps pass cleanly (no errors, no warnings from `trunk check`, tests green, size OK).
   - If `trunk check` suggests auto-fixes (e.g. formatting), apply them first and re-validate.
   - Never bypass this — it's required to maintain SDK stability, footprint, and public API quality.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -28,6 +28,7 @@ You are a senior iOS SDK engineer specializing in stable, lightweight client lib
 - Explicitly dispatch UI work to main actor (`@MainActor`, `DispatchQueue.main.async`).
 - Prefer async/await for new code; existing networking uses completion handlers.
 - Measure & report size impact before proposing dependency or asset changes.
+- Import every module whose members you use, in the file that uses them. `Rokt_Widget` compiles with `MemberImportVisibility`, so relying on a type that reaches us only through a dependency's public API is a compile error rather than a link failure at a consumer.
 
 ### Never automatically (unless specifically requested)
 
--- a/Package.swift
+++ b/Package.swift
@@ -37,7 +37,11 @@ let package = Package(
             path: "Sources/Rokt_Widget",
             resources: [
                 .process("PrivacyInfo.xcprivacy")
-            ]
+            ],
+            // Using a member of a type whose module is not imported in the same
+            // file becomes an error. A type reaching us implicitly through a
+            // dependency's public API is then impossible to use by accident.
+            swiftSettings: [.enableUpcomingFeature("MemberImportVisibility")]
         ),
         .testTarget(
             name: "Rokt_WidgetTests",
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -299,6 +299,31 @@ jobs:
       - name: Run Periphery Scan
         run: periphery scan --format github-actions
 
+  dependency-declaration-check:
+    name: Dependency Declaration Check
+    runs-on: macos-latest-xlarge
+    needs: trunk-check
+    timeout-minutes: 20
+    steps:
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          persist-credentials: false
+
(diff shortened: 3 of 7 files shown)


AGENTS.md
```diff
@@ -90,6 +89,7 @@ You are a senior iOS SDK engineer specializing in stable, lightweight client lib
   3. Run unit tests: Use the `rokt-Example` scheme → Command + U (or `xcodebuild test ...`).
   4. Run periphery scan: `periphery scan` — ensure no unused code is introduced.
   5. If change affects code, assets, or dependencies: Run size report → `Tests/SizeReport/measure_size.sh` and confirm no unacceptable increase.
+  6. If the change touches imports, dependencies, or `Package.swift`: run `./scripts/check_declared_dependencies.sh` — fails when `Rokt_Widget` uses a module that reaches it only through a dependency's public API and is not declared in `Package.swift`.
```
thomson-t: How about using https://docs.swift.org/compiler/documentation/diagnostics/member-import-visibility/ 
It is an upcoming swift compiler feature, but if we enable it, it should flag a build error in these cases.

Package.swift
```diff
@@ -37,7 +37,11 @@ let package = Package(
             path: "Sources/Rokt_Widget",
             resources: [
                 .process("PrivacyInfo.xcprivacy")
-            ]
+            ],
+            // Using a member of a type whose module is not imported in the same
+            // file becomes an error. A type reaching us implicitly through a
+            // dependency's public API is then impossible to use by accident.
+            swiftSettings: [.enableUpcomingFeature("MemberImportVisibility")]
```
thomson-t: We can add the same flag in podspec as well to catch the issue
```
s.pod_target_xcconfig = {
    'OTHER_SWIFT_FLAGS' => '$(inherited) -enable-upcoming-feature MemberImportVisibility'
  }
```

thomson-t (dismissed):

Package.swift
nickolas-dimitrakas: Done in [9d7a1cf](https://github.com/ROKT/rokt-sdk-ios/pull/302/commits/9d7a1cf20e4292b184fb6d61d394688d62876c66)

thomson-t (approved):

